PR #5, opened 2017-06-09
Remove vpn from jenkins

Wynndow:
Openconnect was being used to connect to the GDS VPN so Jenkins could
access github enterprise.

We've moved all relevant repos from enterprise to github, so the vpn is
no longer required.

PR to remove credentilas from jenkins-vars is [here.](https://github.com/alphagov/digitalmarketplace-credentials/pull/1)


Wynndow:
[Get rid of GitHub GDS - M](https://trello.com/c/tqSh5J0J/155-get-rid-of-github-gds-m)

PR #11, opened 2017-07-03
Tag release Docker container with `latest`

allait:
Adding a `latest` tag each time we build a new release container
allows us to pull the set of current images for each app without
checking the current release versions.


Wynndow:
When does the `latest` tag get created? Is it when the image is built with the `-t` flag on line 45?

allait:
@Wynndow yes. The tag is omitted, so `latest` is assigned by default.

Wynndow:
Nice

PR #14, opened 2017-07-24
Create job for taking regular database backups

Wynndow:
This job will deploy the db-backup app, which will create a dump for the live db, zip and encrypt it before uploading to S3.

At the moment it runs against the live db in whichever environment it's built. In future it should create a new postgres service from the latest snapshot taken by PaaS and then taken the dump from there. The logic for this is already in place (see this PR: https://github.com/alphagov/digitalmarketplace-aws/pull/315). This job can be updated when ready.

### DIFF
--- a/job_definitions/database_backup.yml
+++ b/job_definitions/database_backup.yml
@@ -33,9 +33,9 @@
         currentBuild.displayName = "#${BUILD_NUMBER} - ${STAGE}"
           withEnv(["DM_CREDENTIALS_REPO=/var/lib/jenkins/digitalmarketplace-credentials", "CF_HOME=${pwd()}", "STAGE=${STAGE}"]) {
             try {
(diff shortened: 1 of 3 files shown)


job_definitions/database_backup.yml
```diff
@@ -0,0 +1,90 @@
...
+              stage('Login to Cloud Foundry') {
+                paas_credentials = sh(script: '$DM_CREDENTIALS_REPO/sops-wrapper -d $DM_CREDENTIALS_REPO/jenkins-vars/paas_credentials_env.enc', returnStdout: true).trim()
+                withEnv(paas_credentials.tokenize("\n")) {
+                  sh('make paas-login')
+                }
+              }
+
+              stage('Connect to service, create dump and upload to S3') {
+                sh('''
+                  [ -d venv ] || virtualenv venv
```
allait: Why are we not using `make requirements` here?

Wynndow: Because I got in a fight with job builder. I'll change it.

PR #18, opened 2017-08-04
Add job for cleaning and migrating db dumps

Wynndow:
See this PR on the aws repo: [https://github.com/alphagov/digitalmarketplace-aws/pull/317](https://github.com/alphagov/digitalmarketplace-aws/pull/317)
And this one on the scripts repo: [https://github.com/alphagov/digitalmarketplace-scripts/pull/133](https://github.com/alphagov/digitalmarketplace-scripts/pull/133) 

### Add job for migrating and cleaning db dumps
The job will create a temporary postgres service and an app to ssh to it
through.

It will then import the latest production database dump to the new
service, clean it of sensitive data, take a dump of the new database and
then apply that dump to the TARGET_STAGE specified.

It will also upload it to a shared folder on google drive. If the
TARGET_STAGE selected is google-drive, it will not migrate the data to
another stage, just produce the dump in drive.

The default stage is staging, which allows us to set the job to run
every two weeks to automatically update staging with fresh data.

### Index data and update alias
Once data has been migrated to a different stage it needs to be
migrated. There is already a job to do this which we can use. We set the
index name to the days date in keeping with how we currently do it.

Once that index has been created, the alias we use to denote which index
we're using needs switching to it. A script has been created in the
scripts repo to do that and delete old scripts with no aliases.

The next commit will be creating the `update_index_alias` job that's
being refered to here. That job will call that script.

### Add job to update an indexes alias
This job just calls a script which has been added to the scripts repo.
It will update the alias of a given script. If given the
`--delete_old_indexes=yes` flag it will also delete any indexes with no
aliases from that stage.


job_definitions/clean_and_migrate_db_dump.yml
```diff
@@ -0,0 +1,101 @@
...
+    project-type: pipeline
+    description: Takes the latest production database dump, cleans it, and migrates to target stage. Also Google Drive.
+    concurrent: false
+    triggers:
+      - timed: "0 4 * * 0"
+    parameters:
+      - extended-choice:
+          name: TARGET_STAGE
+          type: radio
+          default-value: staging
```
allait: I think default should be google-drive as the least destructive option

job_definitions/clean_and_migrate_db_dump.yml
```diff
@@ -0,0 +1,101 @@
...
+    name: clean-and-migrate-db-dump
+    display-name: Clean and migrate database dump
+    project-type: pipeline
+    description: Takes the latest production database dump, cleans it, and migrates to target stage. Also Google Drive.
+    concurrent: false
+    triggers:
+      - timed: "0 4 * * 0"
+    parameters:
+      - extended-choice:
+          name: TARGET_STAGE
```
allait: Same note as for the script, calling this `TARGET` feels like a way to avoid some of the assumptions built into `STAGE`

job_definitions/clean_and_migrate_db_dump.yml
```diff
@@ -0,0 +1,101 @@
...
+                  returnStdout: true
+                )
+              }
+
+              stage('Migrate data to target stage and google drive') {
+                sh('GDRIVE_EXPORTDATA_FOLDER_ID="{{ jenkins_gdrive_folder_id }}"  make db-cleanup migrate-cleaned-db-dump')
+              }
+
+              stage('Check if migrations required and run'){
+                if (("${TARGET_STAGE}" == 'preview' || "${TARGET_STAGE}" == 'staging') && (env.DUMP_ALEMBIC_VERSION.toInteger() < env.TARGET_STAGE_ALEMBIC_VERSION.toInteger())) {
```
allait: I think it's safe to replace `toInteger` and `<` with `!=`. Migration versions don't have to be numbers, that's just the convention we've been using so far and there's no way for preview or staging to be behind production migration version, so if they're not equal we need to migrate.

job_definitions/clean_and_migrate_db_dump.yml
```diff
@@ -0,0 +1,101 @@
+- job:
+    name: clean-and-migrate-db-dump
```
allait: I wonder if we need a different word for this. Right now we use migrate for both "schema migrations" and "moving data across environments", which is a bit confusing. Restore or apply would work to avoid the clash, but maybe we can find something more distinct.

job_definitions/update_index_alias.yml
```diff
@@ -0,0 +1,49 @@
...
+            stage('Update alias'){
+              git url: 'git@github.com:alphagov/digitalmarketplace-scripts.git', branch: 'update-index-alias', credentialsId: 'github_com_and_enterprise'
+              sh('''
+                VIRTUALENV_ROOT=$([ -z ${VIRTUAL_ENV} ] && echo $(pwd)/venv || echo ${VIRTUAL_ENV})
+                [ -z "${VIRTUAL_ENV}" ] && [ ! -d venv ] && virtualenv venv || true
+                source ${VIRTUALENV_ROOT}/bin/activate
+                ${VIRTUALENV_ROOT}/bin/pip install -r requirements.txt
+
+                if [ $STAGE == 'preview' ]; then
+                  ENDPOINT="search-api.preview.marketplace.team"
```
allait: We usually set things like these either by splitting the job per environment or using Jenkins environment variables (eg $DB_${STAGE}_SEARCH_API_URL) if they're secret. Having them in the job itself feels a bit weird, but if you'd rather keep it as one job then I don't really see a problem with it.

job_definitions/clean_and_migrate_db_dump.yml
Wynndow: I'm going to leave it as staging as we don't have another way to schedule a parameterised job, without creating a another job.

job_definitions/update_index_alias.yml
Wynndow: Nice, have made it in to two jobs and pulled the values from the jenkins variables.

PR #23, opened 2017-08-17
 Add a job to run get-model-data CSV exports

allait:
Script is used to generate data analytics reports and save them as CSVs.

It's using the scripts docker container to run the script and maps a local folder as a volume to keep the reports in the workspace for uploading to Google Drive.

Since Jenkins is used to build the scripts Docker image it should always have the latest version available locally.

Job runs every Monday at 3am UTC.

### Add a job to build a scripts Docker image on each merge

Similar to the app deployment pipeline, we tag the scripts repo with a release tag and run the build image job. Since scripts aren't deployed anywhere that's all this job does.

### Rename build-app job to build-image

Jenkins "build application" job is used to build a Docker image from a tagged repository commit and upload it to Docker Hub.

Up until now it was only used to build application images, but the job isn't really tied to a deployment process, so it can be used to build any docker image. Renaming it to make this clear.

### DIFF
--- a/job_definitions/export_data.yml
+++ b/job_definitions/export_data.yml
@@ -1,0 +1,36 @@
+{% set environments = ['preview', 'production'] %}
+---
+{% for environment in environments %}
+- job:
+    name: "export-data-{{ environment }}"
+    display-name: "Export API model data as CSV - {{ environment }}"
+    project-type: freestyle
+    description: "Runs the get-model-data.py script and saves the generated CSV files to Google Drive"
+    triggers:
+      - timed: "H 3 * * 1"
+    publishers:
+      - trigger-parameterized-builds:
+          - project: notify-slack
+            condition: UNSTABLE_OR_WORSE
+            predefined-parameters: |
+              USERNAME=export-data
+              JOB=Export data {{ environment }}
+              ICON=:foi:
+              STAGE={{ environment }}
+              STATUS=FAILED
+              URL=<${BUILD_URL}consoleFull|#${BUILD_NUMBER}>
+              CHANNEL=#dm-release
+    builders:
+      - shell: |
+
+          # We need to recreate the directory manually since the one created by Docker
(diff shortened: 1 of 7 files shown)


job_definitions/export_data.yml
```diff
@@ -0,0 +1,35 @@
+{% set environments = ['preview', 'production'] %}
```
Wynndow: Why do we need this for preview?

job_definitions/export_data.yml
```diff
@@ -0,0 +1,35 @@
...
+      - shell: |
+
+          # We need to recreate the directory manually since the one created by Docker
+          # when mounting the volume will otherwise be owned by the root user.
+          rm -rf ./data && mkdir data
+
+          docker run --user $(id -u) --volume $(pwd)/data:/app/data digitalmarketplace/scripts scripts/get-model-data.py '{{ environment }}' "$DM_DATA_API_TOKEN_{{ environment|upper }}"
+
+          {% if environment == "production" %}
+          /usr/local/bin/gdrive --config "/var/lib/jenkins/.gdrive" sync upload --delete-extraneous ./data "{{ jenkins_gdrive_csv_data_export_folder_id }}"{% endif %}
```
Wynndow: Can the `{% endif %}` drop down a line? Little confusing style.

job_definitions/export_data.yml
allait: Mostly to make it easier to test. It doesn't upload files anywhere, but you can still check the CSVs from the workspace directory.

job_definitions/export_data.yml
allait: I ran into an annoying Jinja behaviour where it indented all other jobs in the for loop, so that it only ended up creating the production one in the end. But I think it's the empty line after `{% endif %}` that fixes it, not the endif indentation itself, so I'll move it down.

PR #25, opened 2017-08-21
Rename default G-Cloud index to g-cloud-9

samuelhwilliams:
## Summary
Rename default index alias for G-Cloud services to the oldest live framework, in this case g-cloud-9. Pull out environments/default index/default framework to shared file for two jenkins jobs.

### DIFF
--- a/job_definitions/index_services.yml
+++ b/job_definitions/index_services.yml
@@ -1,6 +1,4 @@
-{% set environments = [('preview', 'g-cloud-9'), ('staging', 'g-cloud-9'), ('production', 'g-cloud-9')] %}
----
-{% for environment, frameworks in environments %}
+{% for environment in ['preview', 'staging', 'production'] %}
 - job:
     name: "index-services-{{ environment }}"
     display-name: "Index services - {{ environment }}"
@@ -9,11 +7,11 @@
     parameters:
       - string:
           name: INDEX
-          default: g-cloud
-          description: "Index name to use (eg 'g-cloud-2017-05-22')"
+          default: {{ search_config[environment].default_index }}
+          description: "Index name to use (eg 'g-cloud-9-2017-05-22')"
       - string:
           name: FRAMEWORKS
-          default: {{ frameworks }}
+          default: {{ search_config[environment].frameworks }}
           description: "Comma-separated list of framework slugs that should be indexed (eg 'g-cloud-7,g-cloud-8'). If no frameworks are specified then all currently published services will be indexed."
     triggers:
       - timed: "H 2 * * *"
--- a/job_definitions/update_index_alias.yml
+++ b/job_definitions/update_index_alias.yml
@@ -1,6 +1,4 @@
-{% set environments = ['preview', 'staging', 'production'] %}
----
-{% for environment in environments %}
+{% for environment in ['preview', 'staging', 'production'] %}
 - job:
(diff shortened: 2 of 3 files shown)


TheDoubleK:
...why?

Will this mean I need to re-index locally if I want things to continue to work as expected?

job_definitions/update_index_alias.yml
```diff
@@ -10,7 +10,7 @@
     parameters:
       - string:
           name: ALIAS
-          default: g-cloud
+          default: g-cloud,g-cloud-9
```
TheDoubleK: Do comma-separated aliases in this list get translated into two separate elasticsearch aliases?

samuelhwilliams: This was left in by accident.

samuelhwilliams:
We're doing this so that the Search API URL reflects the framework being searched. Yes, eventually you'll need to reindex when we go down this route and change the searchapi/buyer-frontend to check the specific framework index alias (e.g. `g-cloud-9` rather than `g-cloud`).

job_definitions/search_environment.yml
```diff
@@ -0,0 +1 @@
+{% set environments = [('preview', 'g-cloud-9', 'g-cloud-9'), ('staging', 'g-cloud-9', 'g-cloud-9'), ('production', 'g-cloud-9', 'g-cloud-9')] %}
```
galund: This'd better have a comment saying what each thing is, but otherwise 👍

allait: If you just want to define a variable you can add this to `playbooks/roles/jenkins/defaults/main.yml` as that's the main vars file we use for non-secret stuff. And maybe make this a list of dictionaries.

job_definitions/update_index_alias.yml
```diff
@@ -1,6 +1,6 @@
-{% set environments = ['preview', 'staging', 'production'] %}
+{% include 'search_environment.yml' %}
 ---
-{% for environment in environments %}
+{% for environment, default_index_name, frameworks in environments %}
```
allait: Only `environment` seems to be used, is that what we want?

job_definitions/index_services.yml
```diff
@@ -9,8 +9,8 @@
     parameters:
       - string:
           name: INDEX
-          default: g-cloud
-          description: "Index name to use (eg 'g-cloud-2017-05-22')"
+          default: {{ default_index_name }}
```
allait: Are we planning to switch the search-api to look in `g-cloud-9` instead of `g-cloud`?

allait: Because otherwise as far as I understand this job will just update unused index and the old `g-cloud` one will stay frozen. Or am I missing something?

galund: The search-api already takes the index name as a parameter, so the change we're planning to make is actually on the buyer-frontend. But essentially, yes.

The logic for doing this is explained a bit more in comments on <https://trello.com/c/dhIpjg15/618-data-modelling-documentation>

In summary, we need to be able to derive a framework name given a search index name, and the simplest way to do that is to guarantee that the index is named for (one of) the framework(s) it relates to.

job_definitions/search_environment.yml
samuelhwilliams: Nice thanks, I'll update this tomorrow morning.

PR #34, opened 2017-09-20
Brief Award notifications (4-week and 8-week)

katstevens:
Ticket: https://trello.com/c/poIdAQc0/751-3-create-script-to-send-emails-to-buyers-with-recently-closed-briefs-at-scheduled-intervals

- A template to create 2 jobs, to run the `notify-buyers-to-award-closed-briefs.py` script either 4 weeks or 8 weeks after a brief has closed.
- The jobs should be scheduled for between 8am and 9am (I've used the dynamic scheduler here)
- Both jobs use the `NOTIFY_API_TOKEN` from the jenkins-env credentials

Tested this locally to create 'preview' versions, seems to work ok:
- https://ci.marketplace.team/job/notify-buyers-to-award-closed-briefs-4-weeks-preview/
- https://ci.marketplace.team/job/notify-buyers-to-award-closed-briefs-8-weeks-preview/

Note: the `make jenkins` task isn't keen on indented Jinja tags as the yaml parser complains, so unfortunately there's no indenting for the nested for loops 😿 

Script PR: https://github.com/alphagov/digitalmarketplace-scripts/pull/152

### DIFF
--- a/job_definitions/notify_buyers_to_award_closed_briefs.yml
+++ b/job_definitions/notify_buyers_to_award_closed_briefs.yml
@@ -54,11 +54,6 @@
               CHANNEL=#dm-release
     builders:
       - shell: |
-          [ -d venv ] || virtualenv venv
-
-          . ./venv/bin/activate
-          pip install -r requirements.txt
-
           if [ -n "$OFFSET_DAYS" ]; then
             FLAGS="--offset-days=$OFFSET_DAYS"
           fi
@@ -75,6 +70,6 @@
             FLAGS="$FLAGS --buyer-ids=$BUYER_IDS"
           fi
 
-          ./scripts/notify-buyers-to-award-closed-briefs.py {{ environment }} $DM_DATA_API_TOKEN_{{ environment|upper }} $NOTIFY_API_TOKEN $NOTIFY_TEMPLATE_ID $FLAGS
+          docker run digitalmarketplace/scripts scripts/notify-buyers-to-award-closed-briefs.py {{ environment }} $DM_DATA_API_TOKEN_{{ environment|upper }} $NOTIFY_API_TOKEN $NOTIFY_TEMPLATE_ID $FLAGS
 {% endfor %}
 {% endfor %}
--- a/job_definitions/notify_buyers_to_award_closed_briefs.yml
+++ b/job_definitions/notify_buyers_to_award_closed_briefs.yml
@@ -1,0 +1,80 @@
+{% set environments = ['production'] %}
+{% set reminders = [
+    {"weeks": "4", "notify_template_id": "204d2bf5-5dd3-44c8-a68f-9111e040e831"},
+    {"weeks": "8", "notify_template_id": "bc1f93a8-9529-48b6-9a82-af61bc6da332"}
+  ]
+%}
+---
+{% for environment in environments %}
+{% for reminder in reminders %}
+- job:
+    name: "notify-buyers-to-award-closed-briefs-{{ reminder.weeks }}-weeks-{{ environment }}"
+    display-name: "Notify buyers to award closed briefs (after {{ reminder.weeks }} weeks) - {{ environment }}"
+    project-type: freestyle
+    description: "Send email notification to buyer users to award closed briefs, after {{ reminder.weeks }} weeks."
+    parameters:
+      - string:
+          name: NOTIFY_TEMPLATE_ID
+          default: "{{ reminder.notify_template_id }}"
+          description: "Notify template ID for {{ reminder.weeks }}-week email"
+      - string:
+          name: OFFSET_DAYS
+          default: "{{ reminder.weeks * 7 }}"
+          description: "Notify buyers for briefs that closed X days ago"
+      - bool:
+          name: DRY_RUN
+          default: false
+          description: "List notifications that would be sent without sending the emails"
+      - string:
+          name: DATE_CLOSED
+          description: "Notify buyers for briefs on a specific closing date (YYYY-MM-DD) - must be earlier than the offset days (28 or 56)"
+      - string:
+          name: BUYER_IDS
+          description: "List of buyer IDs to be emailed e.g. 3,4,6"
+    scm:
+      - git:
+          url: git@github.com:alphagov/digitalmarketplace-scripts.git
+          credentials-id: github_com_and_enterprise
+          branches:
+            - master
+          wipe-workspace: false
+    triggers:
+      - timed: "H 8 * * *"
+    publishers:
+      - trigger-parameterized-builds:
+          - project: notify-slack
+            condition: UNSTABLE_OR_WORSE
+            predefined-parameters: |
+              USERNAME=award-closed-briefs-{{ reminder.weeks }}-weeks
+              JOB=Notify buyers to award closed briefs after {{ reminder.weeks }} weeks {{ environment }}
+              ICON=:spaghetti:
+              STAGE={{ environment }}
+              STATUS=FAILED
+              URL=<${BUILD_URL}consoleFull|#${BUILD_NUMBER}>
+              CHANNEL=#dm-release
+    builders:
+      - shell: |
+          [ -d venv ] || virtualenv venv
+
+          . ./venv/bin/activate
+          pip install -r requirements.txt
+
+          if [ -n "$OFFSET_DAYS" ]; then
+            FLAGS="--offset-days=$OFFSET_DAYS"
+          fi
+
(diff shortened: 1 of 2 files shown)


job_definitions/notify_buyers_to_award_closed_briefs_4_weeks.yml
```diff
@@ -0,0 +1,73 @@
+{% set environments = ['production'] %}
+---
+{% for environment in environments %}
+- job:
+    name: "notify-buyers-to-award-closed-briefs-4-weeks-{{ environment }}"
+    display-name: "Notify buyers to award closed briefs (4 weeks) - {{ environment }}"
+    project-type: freestyle
+    description: "Send first email notification to buyer users to award closed briefs, after 4 weeks."
+    parameters:
+      - string:
```
idavidmcdonald: If we change param orders in our script, don't forget to also change them in here

job_definitions/notify_buyers_to_award_closed_briefs_4_weeks.yml
```diff
@@ -0,0 +1,73 @@
...
+    triggers:
+      - timed: "H 8 * * *"
+    publishers:
+      - trigger-parameterized-builds:
+          - project: notify-slack
+            condition: UNSTABLE_OR_WORSE
+            predefined-parameters: |
+              USERNAME=award-closed-briefs-4-weeks
+              JOB=Notify buyers to award closed briefs after 4 weeks {{ environment }}
+              ICON=:question:
```
idavidmcdonald: This is the slack icon that appears when the job fails. We are already using the :question: for the question and answers jenkins job. Not that we need to have different icons for each... but this is your chance to unleash your creative side and pick an appropriate emoji...

idavidmcdonald:
Just back on my further point about compressing the job into one using a for loop, I've now spotted that a few variables might change (4/8 weeks and also the template ID). So maybe give it a try but if it does struggle with readability then feel free to keep it this way.

katstevens:
Yeah, the different template IDs made me hesitate about combining them (also - what if we want to change one of the job to be scheduled differently in future/have a different icon etc?)

idavidmcdonald:
YAGNI I'd say :)

Icon we don't really care about. If the jobs did need different scheduling then we can easily split them but for the moment we should assume we won't need to handle those cases.

job_definitions/notify_buyers_to_award_closed_briefs_4_weeks.yml
katstevens: Probably not appropriate icons but I don't think they've been used by anyone else yet 😸

job_definitions/notify_buyers_to_award_closed_briefs_4_weeks.yml
katstevens: Done

job_definitions/notify_buyers_to_award_closed_briefs.yml
```diff
@@ -0,0 +1,81 @@
+{% set environments = ['production'] %}
+---
+{% for environment in environments %}
+{% with vars = [
```
samuelhwilliams: Maybe this could have a better name than `vars`/`v`. Also any reason you opted to use `with` as opposed to the `set` above?

job_definitions/notify_buyers_to_award_closed_briefs.yml
```diff
@@ -0,0 +1,81 @@
...
+    {"offset": "28", "weeks": "4", "notify_template_id": "204d2bf5-5dd3-44c8-a68f-9111e040e831"},
+    {"offset": "56", "weeks": "8", "notify_template_id": "bc1f93a8-9529-48b6-9a82-af61bc6da332"}
+  ]
+%}
+{% for v in vars %}
+- job:
+    name: "notify-buyers-to-award-closed-briefs-{{ v.weeks }}-weeks-{{ environment }}"
+    display-name: "Notify buyers to award closed briefs ({{ v.weeks }} weeks) - {{ environment }}"
```
samuelhwilliams: Maybe `(after {{ v.weeks }} weeks)` would be clearer?

job_definitions/notify_buyers_to_award_closed_briefs.yml
```diff
@@ -0,0 +1,81 @@
+{% set environments = ['production'] %}
+---
+{% for environment in environments %}
+{% with vars = [
+    {"offset": "28", "weeks": "4", "notify_template_id": "204d2bf5-5dd3-44c8-a68f-9111e040e831"},
+    {"offset": "56", "weeks": "8", "notify_template_id": "bc1f93a8-9529-48b6-9a82-af61bc6da332"}
```
samuelhwilliams: Do we need `offset` here? It feels cleaner to calculate it in the default as `(weeks * 7)`.

job_definitions/notify_buyers_to_award_closed_briefs.yml
```diff
@@ -0,0 +1,81 @@
...
+
+          if [ -n "$DATE_CLOSED" ]; then
+            FLAGS="$FLAGS --date-closed=$DATE_CLOSED"
+          fi
+
+          if [ -n "$BUYER_IDS" ]; then
+            FLAGS="$FLAGS --buyer-ids=$BUYER_IDS"
+          fi
+
+          ./scripts/notify-buyers-to-award-closed-briefs.py {{ environment }} $DM_DATA_API_TOKEN_{{ environment|upper }} $NOTIFY_API_TOKEN $NOTIFY_TEMPLATE_ID $FLAGS
```
samuelhwilliams: Can we use clear argument names here? e.g. `./scripts/notify-buyers-to-award-closed-briefs.py --stage={{ environment }} --api-token=$DM_DATA_API_TOKEN_{{ environment|upper }} --govuk-notify-api_key=$NOTIFY_API_TOKEN --govuk_notify_template_id =$NOTIFY_TEMPLATE_ID $FLAGS`

job_definitions/notify_buyers_to_award_closed_briefs.yml
katstevens: Ooops, yes - David suggested that and I completely forgot 😸

job_definitions/notify_buyers_to_award_closed_briefs.yml
katstevens: I couldn't think of a better name! `jobs` seems a bit vague, how about `reminders`? 

Re: `with`: no reason at all! I will change to `set` for consistency.

job_definitions/notify_buyers_to_award_closed_briefs.yml
katstevens: Not with the script in its current state - docopt throws an error. We could change the script itself, but the `<stage> <api_key> ...` etc is consistent with what we do in other scripts.

samuelhwilliams: Eh, ok. (I really don't like docopt 😛)

job_definitions/notify_buyers_to_award_closed_briefs.yml
```diff
@@ -0,0 +1,80 @@
...
+              STAGE={{ environment }}
+              STATUS=FAILED
+              URL=<${BUILD_URL}consoleFull|#${BUILD_NUMBER}>
+              CHANNEL=#dm-release
+    builders:
+      - shell: |
+          [ -d venv ] || virtualenv venv
+
+          . ./venv/bin/activate
+          pip install -r requirements.txt
```
samuelhwilliams: No longer required if using docker.

katstevens: Good point!

katstevens: I might add a brief note about this on the README.

job_definitions/notify_buyers_to_award_closed_briefs.yml
```diff
@@ -0,0 +1,75 @@
...
+              STATUS=FAILED
+              URL=<${BUILD_URL}consoleFull|#${BUILD_NUMBER}>
+              CHANNEL=#dm-release
+    builders:
+      - shell: |
+          if [ -n "$OFFSET_DAYS" ]; then
+            FLAGS="--offset-days=$OFFSET_DAYS"
+          fi
+
+          if [ "$DRY_RUN" = "true" ]; then
```
samuelhwilliams: This check is probably safer if it checks against `false`, since it's better to assume a dry-run than not. This would do a 'wet' run if someone entered `True`.

PR #49, opened 2017-11-24
Adds job definition for visual regression

MuriloDalRi:
Ticket: https://trello.com/c/y9YBmNH6/41-mvp-implement-visual-regression

### DIFF
--- a/job_definitions/release_pipeline.yml
+++ b/job_definitions/release_pipeline.yml
@@ -12,6 +12,8 @@
     pipeline:
       script: |
 
+        import jenkins.model.Jenkins
+
         def tag_release(tag) {
           try {
             sh("git config user.name 'Jenkins'")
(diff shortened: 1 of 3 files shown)


job_definitions/release_pipeline.yml
```diff
@@ -59,6 +61,81 @@
...
+                    echo "Approving visual regression tests failed: ${inside_error}"
+                    return false
+                  }
+
+                  return true
+                } else {
+                  input(message: "Visual regression tests have run elsewhere since this pipeline triggered the job. They will be re-run here when you proceed.")
+                  return false
+                }
+              } else if (env.FAILED_TEST_ACTION == 'Bypass') {
```
Wynndow: When would this step every be useful?

job_definitions/release_pipeline.yml
```diff
@@ -59,6 +61,81 @@
...
+                    ]
+                  ]
+                }
+
+                echo "Action chosen was ${env.FAILED_TEST_ACTION}"
+              }
+
+              if (env.FAILED_TEST_ACTION == 'Retry') {
+                echo "Retrying visual regression tests on ${stage}."
+                return false
```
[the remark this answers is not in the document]
samuelhwilliams: Yep, it kicks back to the start of the waitUntil loop.

job_definitions/release_pipeline.yml
samuelhwilliams: I don't have a specific use case in mind, but we have the same option on the functional tests, so presumably there's a use case somewhere.

Wynndow: Yeah, I guess having the flexibility is a good thing!

samuelhwilliams: One possible example: someone's testing a framework in a different state (e.g. g-cloud 8 open), and the only changes failing a report are from the banners that appear on the homepage/dashboards. In this case, the framework might be open for an indeterminate amount of time, so you can't really wait to re-run them, but you can't approve the tests either. Bypassing them at that point might be OK as long as the developer was sure no unintended changes slipped through.

PR #50, opened 2017-11-24
Conditionally create admin users for legacy func tests

Wynndow:
The legacy functional tests require certain admin users to be set up.

The tests themselves don't do this.

If we wipe the preview db and restore from a cleaned version (as we do
with staging) we need to recreate those users. This is an annoying
manual chore.

Luckily we have a script to do this. Running it before the tests means
we don't have to remember to do this in the case of db restore.

### DIFF
--- a/job_definitions/functional_tests_legacy.yml
+++ b/job_definitions/functional_tests_legacy.yml
@@ -85,6 +85,8 @@
           mkdir -p reports
           rm -f screenshot*
 
+          ./scripts/create_users.sh example
+
           bundle exec cucumber features --tags ~@wip {% for tag in job.tags %} --tags {{ tag }} {% endfor %} --color --format html --out reports/index.html --format pretty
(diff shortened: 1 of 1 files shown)


job_definitions/functional_tests_legacy.yml
```diff
@@ -85,6 +85,8 @@
           mkdir -p reports
           rm -f screenshot*
 
+          ./scripts/create_users.sh example
```
idavidmcdonald: I assume it doesn't matter that the environment variables that get set during the `create_users` script include some bonus ones not defined above in this jenkins job?
https://github.com/alphagov/digitalmarketplace-functional-tests/blob/jenkins/scripts/envs/example.sh

Wynndow: Yeah I've had a look and they're harmless.

PR #51, opened 2017-11-29
Updated indexing jobs

galund:
Add new job for indexing briefs overnight, fix up services index job.
     - indexing briefs only to happen on preview for now;
     - services job was broken by script name change;
     - we need separate default parameters for the two jobs;
     - also add parameter to allow for new index creation, under manual control;
     - also update DB dump/import script to rebuild briefs index as well as services index;
     - 'update alias' job now cannot have a default alias/target, as it may be needed for both services and briefs. The description will contain some suggestions.
    
https://trello.com/c/5oIKvS1s/100-search-service-for-briefs-plus-indexing-script

Needless to say I have no idea if this code actually works or not :/

### DIFF
--- a/job_definitions/index_briefs.yml
+++ b/job_definitions/index_briefs.yml
@@ -18,7 +18,6 @@
             no frameworks are specified then all currently published briefs will be indexed.
       - string:
           name: CREATE_WITH_MAPPING
-          default:
           description: >
             Specify a mapping to create the index if it doesn't exist. If it does
             exist, the mapping for the index will be updated. There's no default here
--- a/job_definitions/index_briefs.yml
+++ b/job_definitions/index_briefs.yml
@@ -1,0 +1,44 @@
+{% for environment in ['preview'] %}
+- job:
+    name: "index-briefs-{{ environment }}"
+    display-name: "Index briefs - {{ environment }}"
+    project-type: freestyle
+    description: "This imports all briefs (opportunities) in the current API database into the current Elasticsearch index."
+    parameters:
+      - string:
+          name: INDEX
+          default: {{ search_config['briefs'][environment].default_index }}
+          description: "Index name or alias to use (eg 'briefs-digital-outcomes-and-specialists')"
+      - string:
+          name: FRAMEWORKS
+          default: {{ search_config['briefs'][environment].frameworks }}
+          description: >
+            Comma-separated list of framework slugs that should be indexed
+            (eg 'digital-outcomes-and-specialists,digital-outcomes-and-specialists-2'). If
+            no frameworks are specified then all currently published briefs will be indexed.
+      - string:
+          name: CREATE_WITH_MAPPING
+          default:
+          description: >
+            Specify a mapping to create the index if it doesn't exist. If it does
+            exist, the mapping for the index will be updated. There's no default here
+            because using potentially-changed mappings should be done under manual control,
+            on a Production environment.
+    triggers:
+      - timed: "H 3 * * *"
+    publishers:
+      - trigger-parameterized-builds:
+          - project: notify-slack
+            condition: UNSTABLE_OR_WORSE
+            predefined-parameters: |
+              USERNAME=index-briefs
+              JOB=Index briefs {{ environment }} :jenkins_parrot:
+              ICON=:alarm_clock:
+              STAGE={{ environment }}
+              STATUS=FAILED
+              URL=<${BUILD_URL}consoleFull|#${BUILD_NUMBER}>
+              CHANNEL=#dm-release
+    builders:
+      - shell: |
+          docker run digitalmarketplace/scripts scripts/index-to-search-service.py briefs '{{ environment }}' --search-api-token="$DM_SEARCH_API_TOKEN_{{ environment|upper }}" --api-token="$DM_DATA_API_TOKEN_{{ environment|upper }}" --serial --index="${INDEX}" --frameworks="${FRAMEWORKS}" --create-with-mapping="${CREATE_WITH_MAPPING}"
+{% endfor %}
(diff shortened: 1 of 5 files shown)


Wynndow:
Might be worth adding the new briefs job to the list of data jobs in `playbooks/roles/jenkins/defaults/main.yml`. Line 78ish probably.

job_definitions/index_briefs.yml
```diff
@@ -0,0 +1,44 @@
...
+- job:
+    name: "index-briefs-{{ environment }}"
+    display-name: "Index briefs - {{ environment }}"
+    project-type: freestyle
+    description: "This imports all briefs (opportunities) in the current API database into the current Elasticsearch index."
+    parameters:
+      - string:
+          name: INDEX
+          default: {{ search_config[environment].default_index }}
+          description: "Index name or alias to use (eg 'briefs-digital-outcomes-and-specialists-2')"
```
Wynndow: We dropped the `-2` off the end of the index name

job_definitions/index_briefs.yml
```diff
@@ -0,0 +1,44 @@
...
+      - string:
+          name: FRAMEWORKS
+          default: {{ search_config[environment].frameworks }}
+          description: >
+            Comma-separated list of framework slugs that should be indexed
+            (eg 'digital-outcomes-and-specialists-1,digital-outcomes-and-specialists-2'). If
```
Wynndow: it's just `digital-outcomes-and-specialists`, no `-1`

job_definitions/index_briefs.yml
samuelhwilliams: Guess I'm not up-to-date on this; is the decision, then, that multiple framework iterations are potentially going to be stored in a single index under different document names?

galund: They definitely will in the case of DOS - the existing list of opportunities includes (if you choose to search for closed briefs) the DOS-1 briefs, too.

job_definitions/index_briefs.yml
```diff
@@ -0,0 +1,44 @@
...
+            no frameworks are specified then all currently published briefs will be indexed.
+      - string:
+          name: CREATE_WITH_MAPPING
+          default:
+          description: >
+            Specify a mapping to create the index if it doesn't exist. If it does
+            exist, the mapping for the index will be updated. Don't specify this
+            option when running from a batch job, as both creating indexes and updating
```
samuelhwilliams: `batch job` should maybe say regular/automated job? Running this at all is presumably a batch job in that it indexes a batch of services (unless my nomenclature is off).

If this option should never be provided to the job at all, then we should just not offer it at all?

galund: Yeah, I had trouble with the wording here. It is intended that you can use the option when running manually, and Jenkins is supposed to be the easy way to do that (a fact I forgot when writing the docopt in the script).

I think that technically 'batch' refers to the fact that the _jobs_ are batched up to run (i.e. automatically), but seeing as it's definitely confusing, let's try another word instead. Of course, even the manual jobs are automated on some level...

How about "This option should only be used when running under manual control (i.e. never from an automated/timed job), as...".

"Timed" is the word that we use in the "triggers" section of the job definition, anyway.

job_definitions/index_services.yml
```diff
@@ -21,13 +28,13 @@
             condition: UNSTABLE_OR_WORSE
             predefined-parameters: |
               USERNAME=index-services
-              JOB=Index services {{ environment }}
+              JOB=Index services {{ environment }} :jenkins_parrot:
```
idavidmcdonald: We don't usually put emoji's in the job name (the icon is the thing we usually change). Was that the intent?

job_definitions/index_services.yml
```diff
@@ -7,12 +7,19 @@
...
+          description: "Index name or alias to use (eg 'g-cloud-9-2017-05-22')"
       - string:
           name: FRAMEWORKS
-          default: {{ search_config[environment].frameworks }}
+          default: {{ search_config['services'][environment].frameworks }}
           description: "Comma-separated list of framework slugs that should be indexed (eg 'g-cloud-7,g-cloud-8'). If no frameworks are specified then all currently published services will be indexed."
+      - string:
+          name: CREATE_WITH_MAPPING
+          description: >
+            Specify a mapping to create the index if it doesn't exist. If it does
```
idavidmcdonald: I find this description a little hard to understand (it's probably because I know nothing about mappings). I don't quite get what you mean by 'under manual control'.

job_definitions/index_briefs.yml
```diff
@@ -0,0 +1,43 @@
+{% for environment in ['preview'] %}
+- job:
+    name: "index-briefs-{{ environment }}"
+    display-name: "Index briefs - {{ environment }}"
+    project-type: freestyle
+    description: "This imports all briefs (opportunities) in the current API database into the current Elasticsearch index."
+    parameters:
+      - string:
+          name: INDEX
+          default: {{ search_config['briefs'][environment].default_index }}
```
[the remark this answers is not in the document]
idavidmcdonald: Note, this isn't the only place it's being done in this PR.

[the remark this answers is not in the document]
galund: Yes, so these are references to keys in `playbooks/roles/jenkins/defaults/main.yml` (see `search_config` in that file).

The `'briefs'` thing is a constant. `environment` is parameterized, i.e. it's a jinja variable defined in a loop around the entire job. Which is a bit weird :-)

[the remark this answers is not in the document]
idavidmcdonald: Gotcha. Ta

PR #56, opened 2017-11-30
Add admin-manager and admin-framework-manager users to job configs

benvand:
Make credentials available to functional tests, smoke tests and visual regression run.


samuelhwilliams:
Or maybe this PR has nothing to do with the one Kev put up on the VR tests repo, in which case maybe it doesn't matter. Although nothing would be using these vars, so should we be passing them in?

benvand:
Looks like those were added to master by @TheDoubleK on Friday

PR #57, opened 2017-11-30
Add visual regression to build monitor

TheDoubleK:
Does what it says on the tin.


TheDoubleK:
I tried ordering things but it just puts them in alphabetical order no matter what order you put in the file. I didn't dig any deeper though, and there may be a way. But adding things in the Jenkins UI it just gives you a checklist of projects and no way to order them.

TheDoubleK:
Let's get this in, but maybe team needs to discuss what is essential for the board. This one feels like it is because it should block people from merging new stuff. The email jobs I care less about on the board; the slack notifications might be enough.

PR #61, opened 2017-12-12
Update some notify jobs to use docker/py3

risicle:
Story https://trello.com/c/pCNtBX6A/108-converting-jenkins-jobs-to-python-3

Also environment-parametrized `notify_suppliers_of_dos_opportunities`: though we're only actually adding a job for the "production" environment here, it is occasionally useful to be able to spawn up a job that will draw its data from preview or staging.

Omitting `update_index_alias` for now as it looks like that will need a few changes to the script to be able to operate without requiring the presence of `digitalmarketplace-credentials` and all its dependencies. Or the answer might be to make _all_ the scripts depend on `digitalmarketplace-credentials`. Who knows.

All of these have been pushed to jenkins with positive results.

### DIFF
--- a/job_definitions/notify_buyers_when_requirements_close.yml
+++ b/job_definitions/notify_buyers_when_requirements_close.yml
@@ -37,11 +37,6 @@
               CHANNEL=#dm-release
     builders:
       - shell: |
-          [ -d venv ] || virtualenv venv
-
-          . ./venv/bin/activate
-          pip install -r requirements.txt
-
           if [ -n "$DATE_CLOSED" ]; then
             FLAGS="--date-closed=$DATE_CLOSED"
           fi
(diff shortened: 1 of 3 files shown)


galund:
`update_index_alias` doesn't need access to anything that isn't already required by `index_services`, so I'd have thought this would be feasible.

Happy to do it myself but maybe we should catch up in person first in case I misunderstood something.

risicle:
Yup in theory it's straightforward but as I said, the way each different script wants to take its credentials is different and in `update-index-alias`'s case there is no way of specifying auth tokens other than pointing the script to the credentials repo. Which then means we have a(n undeclared) dependency on all its dependencies (`aws-auth`, `jq`, `sops`...).

galund:
Yup, it might be good to decide if passing on the command line was the normal way, or the other thing was the normal way, and then just stick with one thing or the other. (But other views sought, cc @Wynndow perhaps?)

PR #62, opened 2017-12-13
Reorganise plugins

katstevens:
The plugin list on the Ansible config did not match reality of the dashboard.

- Alphabetise the list so we can easily see missing plugins and update with the missing ones
- Add some info to the README about updating plugins

NB: I haven't checked whether we actually **use** all of these! We should do a proper plugin audit at a later date.

### DIFF
--- a/playbooks/roles/jenkins/defaults/main.yml
+++ b/playbooks/roles/jenkins/defaults/main.yml
@@ -30,9 +30,11 @@ dm_db_applications:
 
 jenkins_plugins:
   - ansicolor
+  - blueocean
   - build-monitor-plugin
   - build-name-setter
   - conditional-buildstep
+  - cvs
   - extended-choice-parameter
   - git
   - github
@@ -43,6 +45,10 @@ jenkins_plugins:
   - nodejs
   - parameterized-trigger
   - postbuildscript
+  - rebuild
+  - translation
+  - workflow-aggregator  # aka 'Pipeline' suite
+  - ws-cleanup
 
 jenkins_config_templates:
   - config.xml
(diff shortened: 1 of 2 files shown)


playbooks/roles/jenkins/defaults/main.yml
```diff
@@ -29,20 +29,26 @@ dm_db_applications:
...
-  - extended-choice-parameter
+  - htmlpublisher
   - lockable-resources
+  - nodejs
+  - parameterized-trigger
+  - postbuildscript
+  - rebuild
+  - translation
+  - workflow-aggregator  # aka 'Pipeline' suite
+  - ws-cleanup
```
samuelhwilliams: I'm 99% sure we don't need ws-cleanup - I think I accidentally installed that on our CI rather than my local instance when we were implementing VR tests. And then forgot to uninstall it. 🙈

katstevens: If you take it off the dashboard then I'll take it off this list 😸

playbooks/roles/jenkins/defaults/main.yml
```diff
@@ -29,20 +29,26 @@ dm_db_applications:
   - api
 
 jenkins_plugins:
+  - ansicolor
+  - blueocean
```
TheDoubleK: We definitely don't use this.

playbooks/roles/jenkins/defaults/main.yml
samuelhwilliams: 😆

katstevens:
OK, I will take the plunge and remove `ws-cleanup` and `blueocean` - we can always put them back in if something goes wrong...

playbooks/roles/jenkins/defaults/main.yml
samuelhwilliams: Just checked; it was installed November 28th, so it was definitely me. Thanks for removing. 👍

PR #65, opened 2018-01-18
Remove redundant virtualenv/pip install from Publish Manual job

katstevens:
The manual's `make html` command now creates and applies its own virtualenv, and installs 
 requirements. See https://github.com/alphagov/digitalmarketplace-manual/pull/46 for details.


katstevens:
Abandoning this idea - it's easier for Jenkins to create/activate/install its own virtualenv.

PR #75, opened 2018-03-06
database_migration_paas job: notify slack of job result

risicle:
https://trello.com/c/2QlnK7kW/197-api-release-pipeline-doesnt-notify-slack-if-the-preview-migration-fails
 
I wonder if this works.

There may be only one way to find out...

### DIFF
--- a/job_definitions/database_migration_paas.yml
+++ b/job_definitions/database_migration_paas.yml
@@ -22,7 +22,23 @@
           description: "Release name (eg 'release-42') to run migrations from"
     pipeline:
       script: |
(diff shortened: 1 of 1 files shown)


risicle:
This new version **does** work. Though yes, the "success" message is a little noisy, but it was useful for testing. Now that I know the `notify_slack` function works I can maybe comment out the success message.

risicle:
Actually nah sod it I'm going to enable the success message. It's not like we do _that_ many api releases anyway.

risicle:
Quack.

PR #77, opened 2018-03-24
Add jobs for rotating tokens

samuelhwilliams:
## Summary
Adds a new pipeline that allows for easily rotating data/search API
tokens in a given environment. Also updates the existing job for
releasing an application to paas to allow building without specifying
a release version. In this case, it will determine the existing release
by finding the app's primary route (via CF CLI) and hitting its
`_status` endpoint to find the current release tag. After doing that, it
will re-release that same version to PaaS, allowing new configuration to
be pulled in. The detection and deployment happens within the specific
app and stage's Jenkins lock, meaning it is effectively atomic and so
asynchronicity should not be a concern.

 ## Ticket
https://trello.com/c/4crGe6F0/169

### DIFF
--- a/job_definitions/release_application_paas.yml
+++ b/job_definitions/release_application_paas.yml
@@ -22,57 +22,73 @@
 {% endfor %}
       - string:
           name: RELEASE_NAME
-          description: "Release name (eg 'release-42') to deploy"
+          description: "Release name (eg 'release-42') to deploy. If blank, the current release will be re-deployed. NOTE: router version cannot be derived - *must* be manually specified."
     pipeline:
       script: |
+        def notify_slack(icon, status) {
+          build job: "notify-slack",
+            parameters: [
+              string(name: 'USERNAME', value: "release-application-to-paas"),
+              string(name: 'ICON', value: icon),
+              string(name: 'JOB', value: "Release ${APPLICATION_NAME} to ${STAGE}"),
+              string(name: 'CHANNEL', value: "#dm-release"),
+              string(name: 'PROJECT', value: "${APPLICATION_NAME}"),
+              text(name: 'STAGE', value: "${STAGE}"),
+              text(name: 'RELEASE_NAME', value: "<https://github.com/alphagov/digitalmarketplace-${APPLICATION_NAME}/tree/${RELEASE_NAME}|${RELEASE_NAME}>"),
+              text(name: 'STATUS', value: status),
+              text(name: 'URL', value: "<${BUILD_URL}consoleFull|${BUILD_DISPLAY_NAME}>")
+            ]
+        }
 
-          def notify_slack(icon, status) {
-            build job: "notify-slack",
-                  parameters: [
-                    string(name: 'USERNAME', value: "release-application-to-paas"),
-                    string(name: 'ICON', value: icon),
-                    string(name: 'JOB', value: "Release ${APPLICATION_NAME} to ${STAGE}"),
-                    string(name: 'CHANNEL', value: "#dm-release"),
-                    string(name: 'PROJECT', value: "${APPLICATION_NAME}"),
-                    text(name: 'STAGE', value: "${STAGE}"),
-                    text(name: 'RELEASE_NAME', value: "<https://github.com/alphagov/digitalmarketplace-${APPLICATION_NAME}/tree/${RELEASE_NAME}|${RELEASE_NAME}>"),
-                    text(name: 'STATUS', value: status),
-                    text(name: 'URL', value: "<${BUILD_URL}consoleFull|${BUILD_DISPLAY_NAME}>")
-                  ]
-          }
+        try {
+          node {
+            git url: 'git@github.com:alphagov/digitalmarketplace-aws.git', branch: 'master', credentialsId: 'github_com_and_enterprise'
 
-          try {
-            node {
-              git url: 'git@github.com:alphagov/digitalmarketplace-aws.git', branch: 'master', credentialsId: 'github_com_and_enterprise'
+            stage('Prepare') {
+              build job: "update-credentials"
+              currentBuild.displayName = "#${BUILD_NUMBER} - ${STAGE} - ${APPLICATION_NAME} - #${RELEASE_NAME}"
+              sh('make paas-clean')
+              sh('pyenv local 3.6.2 && make requirements')
+            }
 
-              stage('Prepare') {
-                  build job: "update-credentials"
-                  currentBuild.displayName = "#${BUILD_NUMBER} - ${STAGE} - ${APPLICATION_NAME} - #${RELEASE_NAME}"
-                  sh('make paas-clean')
-                  sh('pyenv local 3.6.2 && make requirements')
-              }
+            stage('Deploy') {
+              withEnv(["DM_CREDENTIALS_REPO=/var/lib/jenkins/digitalmarketplace-credentials", "CF_HOME=${pwd()}"]) {
+                paas_credentials = sh(script: '$DM_CREDENTIALS_REPO/sops-wrapper -d $DM_CREDENTIALS_REPO/jenkins-vars/paas_credentials_env.enc', returnStdout: true).trim()
+                withEnv(paas_credentials.tokenize("\n")) {
+                  sh('make paas-login')
+                }
 
-              stage('Deploy') {
-                withEnv(["DM_CREDENTIALS_REPO=/var/lib/jenkins/digitalmarketplace-credentials", "CF_HOME=${pwd()}"]) {
-                  paas_credentials = sh(script: '$DM_CREDENTIALS_REPO/sops-wrapper -d $DM_CREDENTIALS_REPO/jenkins-vars/paas_credentials_env.enc', returnStdout: true).trim()
-                  withEnv(paas_credentials.tokenize("\n")) {
-                      sh('make paas-login')
+                lock("paas-${STAGE}-${APPLICATION_NAME}-deploy") {
+                  if ("$RELEASE_NAME" == "" && "$APPLICATION_NAME" != "router") {
+                    echo "No release specified. Detecting currently deployed release version of ${APPLICATION_NAME} ..."
+
+                    env.RELEASE_NAME = sh(script: '''
+                      APP_ROUTES=$(cf curl /v2/apps/\$(cf app ${APPLICATION_NAME} --guid)/routes)
+                      APP_SUBDOMAIN=$(echo \"\$APP_ROUTES\" | jq -crM ".resources[0].entity.host")
+                      APP_PATH=$(echo \"\$APP_ROUTES\" | jq -crM ".resources[0].entity.path")
+                      curl -s "https://\$APP_SUBDOMAIN.cloudapps.digital\$APP_PATH/_status" | jq -crM ".version"
+                      ''',
+                      returnStdout: true
+                    ).trim()
+
+                    echo "Current deployed version of ${APPLICATION_NAME}: ${RELEASE_NAME}"
+                    currentBuild.displayName = "#${BUILD_NUMBER} - ${STAGE} - ${APPLICATION_NAME} - #${RELEASE_NAME}"
                   }
-                  lock("paas-${STAGE}-${APPLICATION_NAME}-deploy") {
-                      sh('make ${STAGE} deploy-app')
-                  }
+
+                  sh("make ${STAGE} deploy-app")
                 }
               }
             }
-
-            notify_slack(':sheep:', 'SUCCESS')
-
-          } catch(err) {
-              currentBuild.result = 'FAILURE'
-              echo "Error: ${err}"
-              notify_slack(':kaboom:', 'FAILED')
-          } finally {
-            node {
-              sh('make paas-clean')
-            }
           }
+
+          notify_slack(':sheep:', 'SUCCESS')
+
+        } catch(err) {
+            currentBuild.result = 'FAILURE'
+            echo "Error: ${err}"
+            notify_slack(':kaboom:', 'FAILED')
+        } finally {
+          node {
+            sh('make paas-clean')
+          }
+        }
--- a/job_definitions/rotate_api_tokens.yml
+++ b/job_definitions/rotate_api_tokens.yml
@@ -1,0 +1,113 @@
+{#
+  Our apps have to be deployed in a specific order when recycling tokens: Search API first, then Data API, then the frontend apps.
+  Any of our frontend apps can talk talk to any of our APIs. The Data API can talk to the Search API. The Search API doesn't talk to anything else.
+  Our APIs receive a list of tokens they should accept. On top of this, apps that need to talk to an API use the _first_ token from the list of tokens as the one they'll use to authenticate.
+  Therefore, any API that receives requests from some other app needs to have their list of accepted tokens updated before any app starts using one of the new tokens for authentication.
+  As the Search API doesn't talk to the Data API, we can update it safely first.
+  After that, the Data API can safely be updated as the Search API will now accept the new token it will receive.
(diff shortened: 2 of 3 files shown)


katstevens:
Are we going to have a Slack notification on success/failure?

samuelhwilliams:
Ah that's a good idea. Yes, let me add that now.

samuelhwilliams:
Added @katstevens 👍

job_definitions/release_application_paas.yml
```diff
@@ -53,14 +53,23 @@
...
+                    }
                   }
-                  lock("paas-${STAGE}-${APPLICATION_NAME}-deploy") {
-                      sh('make ${STAGE} deploy-app')
+
+                lock("paas-${STAGE}-${APPLICATION_NAME}-deploy") {
+                  if ("$RELEASE_NAME" == "") {
+                    echo "No release specified. Detecting currently deployed release version of ${APPLICATION_NAME} ..."
+
+                    RELEASE_NAME = sh(script: "curl -s https://\$(cf curl /v2/apps/\$(cf app ${APPLICATION_NAME} --guid)/routes | jq -crM '.resources[0].entity.host').cloudapps.digital\$(cf curl /v2/apps/\$(cf app ${APPLICATION_NAME} --guid)/routes | jq -crM '.resources[0].entity.path')/_status | jq -crM '.version'", returnStdout: true)
```
katstevens: This line has a lot going on! Let me check if I understand the different parts of what's happening here:
- Get the Cloud Foundry app ID with `cf app ${APPLICATION_NAME} --guid` (twice?)
- Use that app ID to get the resource host and path, via two separate `curl` requests
- Construct the url to get that resource's `/_status`, to get the latest `version`

I think splitting out fetching the app ID into a separate var might make this more readable (that's assuming I've read all this right!)

samuelhwilliams: Added a comment to hopefully clear things up a little.

Painful :(

job_definitions/release_application_paas.yml
```diff
@@ -53,14 +53,33 @@
...
+                      This is a pretty painful command to parse. It essentially does the following:
+
+                      1) Hits `cf app ${APPLICATION_NAME} --guid` to get the CloudFoundry GUID of an application.
+                      2) Uses `cf curl /v2/apps/guid/routes` to work out the subdomain/path suffix an application is registered on.
+                      3) Uses `curl https://host.cloudapps.digital/path/_status` to retrieve the currently-released version of the app.
+
+                      We didn't find a way to do this across multiple lines, which would allow having intermediate variables to increase
+                      readability, due to how Jenkins escapes/injects variables between Groovy and the subshell.
+                    */
+                    RELEASE_NAME = sh(script: "curl -s https://\$(cf curl /v2/apps/\$(cf app ${APPLICATION_NAME} --guid)/routes | jq -crM '.resources[0].entity.host').cloudapps.digital\$(cf curl /v2/apps/\$(cf app ${APPLICATION_NAME} --guid)/routes | jq -crM '.resources[0].entity.path')/_status | jq -crM '.version'", returnStdout: true)
```
katstevens: George spotted this in the index alias script - multi line script within a single `sh()` that might work? https://github.com/alphagov/digitalmarketplace-jenkins/blob/master/job_definitions/update_index_alias.yml#L31

job_definitions/update_api_tokens.yml
```diff
@@ -0,0 +1,74 @@
+{% set stages = [{'name': 'Add new tokens', 'command': 'add-new'}, {'name': 'Remove old tokens', 'command': 'remove-old'}] %}
+{% set app_groups = [['search-api'], ['api'], ['admin-frontend', 'briefs-frontend', 'brief-responses-frontend', 'buyer-frontend', 'supplier-frontend', 'user-frontend']] %}
+---
+- job:
+    name: update-api-tokens
+    display-name: "Update API tokens"
+    project-type: pipeline
+    descript1ion: Update Data and Search API tokens for a given environment.
```
idavidmcdonald: descript1ion?

idavidmcdonald: Could this be slightly more descriptive please i.e. what does 'update' mean

idavidmcdonald: Maybe we could call this 'rotate_api_tokens' instead of update?

job_definitions/release_application_paas.yml
```diff
@@ -53,14 +53,33 @@
...
+                lock("paas-${STAGE}-${APPLICATION_NAME}-deploy") {
+                  if ("$RELEASE_NAME" == "") {
+                    echo "No release specified. Detecting currently deployed release version of ${APPLICATION_NAME} ..."
+
+                    /*
+                      This is a pretty painful command to parse. It essentially does the following:
+
+                      1) Hits `cf app ${APPLICATION_NAME} --guid` to get the CloudFoundry GUID of an application.
+                      2) Uses `cf curl /v2/apps/guid/routes` to work out the subdomain/path suffix an application is registered on.
+                      3) Uses `curl https://host.cloudapps.digital/path/_status` to retrieve the currently-released version of the app.
```
Wynndow: Might be worth changing this to `curl https://<subdomain>.cloudapps.digital/<path>/_status` to make it really clear that those parts are dynamic.

job_definitions/release_application_paas.yml
```diff
@@ -53,14 +53,33 @@
...
+
+                lock("paas-${STAGE}-${APPLICATION_NAME}-deploy") {
+                  if ("$RELEASE_NAME" == "") {
+                    echo "No release specified. Detecting currently deployed release version of ${APPLICATION_NAME} ..."
+
+                    /*
+                      This is a pretty painful command to parse. It essentially does the following:
+
+                      1) Hits `cf app ${APPLICATION_NAME} --guid` to get the CloudFoundry GUID of an application.
+                      2) Uses `cf curl /v2/apps/guid/routes` to work out the subdomain/path suffix an application is registered on.
```
Wynndow: Should this be subdomain or host? It's weird, my understanding is that `cloudapps.digital` is the host and `dm-api-preview`, for example, is the subdomain. But the cf api seems to use 'host' for what I thought was the subdomain.

job_definitions/update_api_tokens.yml
```diff
@@ -0,0 +1,74 @@
...
+            string(name: 'CHANNEL', value: "#dm-release"),
+            text(name: 'STAGE', value: "${STAGE}"),
+            text(name: 'STATUS', value: status),
+            text(name: 'URL', value: "<${BUILD_URL}consoleFull|${BUILD_DISPLAY_NAME}>")
+          ]
+        }
+
+{% for stage in stages %}
+        stage("{{ stage.name }}") {
+{% if loop.index == 1 %}
```
Wynndow: This threw me a bit, why you were using index '1'. Had to look at the docs to find out `loop.index` is 1 indexed. `loop.index()` is 0 indexed. Would it be more intuitive to use `loop.index()`

job_definitions/update_api_tokens.yml
```diff
@@ -0,0 +1,74 @@
...
+          currentBuild.displayName = "#${BUILD_NUMBER} - ${STAGE}"
+{% endif %}
+          node {
+            try {
+              sh('docker run -e GITHUB_ACCESS_TOKEN digitalmarketplace/scripts scripts/update-api-tokens.sh "{{ stage.command }}" "${STAGE}"')
```
Wynndow: Where does `GITHUB_ACCESS_TOKEN` come from? If it's a variable should it be `"$GITHUB_ACCESS_TOKEN"`?

job_definitions/update_api_tokens.yml
```diff
@@ -0,0 +1,74 @@
...
+            }
+          }
+        }
+
+        stage('Deploy') {
+          milestone()
+          input(message: "Do not continue until the Pull Request updating API tokens has been merged to master.")
+          milestone()
+
+{% for app_group in app_groups %}
```
Wynndow: The docs seem to suggest that only stages within a `parallel` block will be run in parallel. This gives you three `parallel` blocks, but the first two only have one stage in them for api and search-api. It probably doesn't do any harm but could be confusing.

Also, what does having `"{{ app }}"` as the top level key give you? In the docs it's just got `stage` as the top level.

job_definitions/update_api_tokens.yml
samuelhwilliams: There's `index` and `index0` I think. Haven't heard of `index()` but if that also exists it's super weird.

job_definitions/update_api_tokens.yml
samuelhwilliams: It's an environment variable. When passing via the `-e` flag of docker run it will look for an env variable called that on the host and inject it, along with its value, into the container. We do this so we don't have to show the token itself in the command.

job_definitions/release_application_paas.yml
samuelhwilliams: 👍

job_definitions/update_api_tokens.yml
samuelhwilliams: 👍

job_definitions/release_application_paas.yml
samuelhwilliams: 👍

job_definitions/update_api_tokens.yml
Wynndow: derp. I read `0` as`()`

job_definitions/update_api_tokens.yml
samuelhwilliams: 👍

job_definitions/update_api_tokens.yml
Wynndow: So it's an env variable on the Jenkins box? I can only find `GITHUB_TOKEN` in the jenkins vars in the creds repo? And it'll still need to be `${GITHUB_TOKEN}`, no?

job_definitions/release_application_paas.yml
samuelhwilliams: Yeah the terminology gets pretty mixed up between reality/CF... Will change later references to subdomain.

job_definitions/update_api_tokens.yml
samuelhwilliams: Hmm, in the docs I looked at you need to give parallel a map, so the app name is just filling that need rather than anything more meaningful. 🤔

Wynndow: I was looking here: https://jenkins.io/blog/2017/09/25/declarative-1/

samuelhwilliams: https://jenkins.io/doc/book/pipeline/jenkinsfile/#parallel-execution

samuelhwilliams: As for why three parallel blocks when the first two are redundant - you're right, it doesn't actually do anything in the case of the search-api and the api. Those will happen consecutively rather than at the same time, but that's what they need to do. It was just a slightly simpler template to group them and have them all run within parallel blocks.

samuelhwilliams: This is the alternative:


```{% for app_group in app_groups %}
{% if app_group|length == 1 %}
          stage("{{ app_group[0] }}") {
            build job: "release-app-paas", parameters: [
              string(name: "STAGE", value: "${STAGE}"),
              string(name: "APPLICATION_NAME", value: "{{ app_group[0] }}")
            ]
          }
{% elif app_group|length > 1 %}
          parallel(
{% for app in app_group %}
            "{{ app }}": {
              stage("{{ app }}") {
                build job: "release-app-paas", parameters: [
                  string(name: "STAGE", value: "${STAGE}"),
                  string(name: "APPLICATION_NAME", value: "{{ app }}")
                ]
              }
            },
{% endfor %}
          )
{% endif %}
{% endfor %}
        }```

Comment (Wynndow) [reply job_definitions/update_api_tokens.yml]: Gross. Just keep the original I think.

File: job_definitions/rotate_api_tokens.yml [hunk 9]
```diff
@@ -0,0 +1,92 @@
...
+                ]
+              }
+            },
+{% endfor %}
+          )
+{% endfor %}
+        }
+
+{% if loop.index == 1 %}
+        stage("Update Jenkins Config") {
```
idavidmcdonald: It feels like as a team we have a poor understanding of why the latest creds need to be in the jenkins repo. I roughly know what it's doing as you showed me the screen earlier but I guess I still don't really know why we have that stuff there and the importance of it being up to date. Could you add something here or in the manual maybe to try and get that knowledge explicit?

job_definitions/rotate_api_tokens.yml
```diff
@@ -0,0 +1,92 @@
...
+                  string(name: "APPLICATION_NAME", value: "{{ app }}")
+                ]
+              }
+            },
+{% endfor %}
+          )
+{% endfor %}
+        }
+
+{% if loop.index == 1 %}
```
idavidmcdonald: On a small note, could we indent our loops so we can see the logic flow based on indentation? (I know we don't always do this, myself included but I think we should change that)

job_definitions/release_application_paas.yml
```diff
@@ -22,57 +22,73 @@
...
-              sh('make paas-clean')
-            }
+        } catch(err) {
+            currentBuild.result = 'FAILURE'
+            echo "Error: ${err}"
+            notify_slack(':kaboom:', 'FAILED')
+        } finally {
+          node {
+            sh('make paas-clean')
           }
```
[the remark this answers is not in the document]
samuelhwilliams: Yeah, if you add `?w=1` to the files URL, GitHub will ignore whitespace and you'll see a more sane diff. I did indeed tweak indentation.

job_definitions/rotate_api_tokens.yml
```diff
@@ -0,0 +1,92 @@
...
+          choices:
+            - preview
+            - staging
+            - production
+    pipeline:
+      script: |
+        def notify_slack(icon, status) {
+          build job: "notify-slack",
+          parameters: [
+            string(name: 'USERNAME', value: "update_api_tokens"),
```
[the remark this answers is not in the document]
samuelhwilliams: 👍

job_definitions/rotate_api_tokens.yml
samuelhwilliams: We generally don't indent them for a reason - jinja preserves the whitespace and it leads to the output being indented all over the place. For things where whitespace matters (e.g. yaml), this makes things break.

idavidmcdonald: Would some of the jinja control characters fix that? https://stackoverflow.com/questions/35775207/remove-unnecessary-whitespace-from-jinja-rendered-template

job_definitions/rotate_api_tokens.yml
samuelhwilliams: We don't really have a section on Jenkins in the manual at all. Feels like a page that could do with quite a bit of info.

We do have stuff scattered throughout the manual on Jenkins, but having the core of it in one place would be good.

job_definitions/rotate_api_tokens.yml
samuelhwilliams: I tried using these before with little success. I can try again if you think it's an isssue.

idavidmcdonald: Don't spend any more than 2 minutes, it's not a game changer

job_definitions/rotate_api_tokens.yml
idavidmcdonald: Great point, I'll add a new page in the manual with all our jenkins stuff on my to do list

samuelhwilliams: It can be a shared todo list. It's a big page so don't take it all on yourself.

PR #78, opened 2018-03-26
Docker / python 3k for index alias job

galund:
- auth tokens are passed in via the environment; depends on PR https://github.com/alphagov/digitalmarketplace-scripts/pull/210

https://trello.com/c/yNVT3P2Y/227-updating-index-alias-script-to-python-3


galund:
I'm totally confused by that chunk of documentation - the distinction between stage and environment seems to be "it's the same as stage" in which case I don't even... huh?

PR #80, opened 2018-03-29
Remove pipeline timeout

samuelhwilliams:
## Summary
Removes the 24-hour pipeline timeout from deployment jobs. This seems to
have only caused problems in reality. I'm not sure what the original
intent was behind the jobs dying but I think it was related to not
having executors blocked by paused jobs. I don't think this is the case,
and if it theoretically were, we'd encounter it anyway when we pause
before deploy to staging.

@Wynndow is there a reason for this we aren't aware of?

### DIFF
--- a/job_definitions/release_pipeline.yml
+++ b/job_definitions/release_pipeline.yml
@@ -34,20 +34,19 @@
               }
             } catch(error) {
               echo "Functional tests failed: ${error}"
-              timeout(time:1, unit:'DAYS') {
-                script {
-                  env.FAILED_TEST_ACTION = input message: 'Functional tests failed. Retry, skip or abort?',
-                    parameters: [
-                      [
-                        $class: 'ChoiceParameterDefinition',
-                        choices: 'Retry\nSkip',
-                        description: 'Failed functional tests action',
-                        name: 'action'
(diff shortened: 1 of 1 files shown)


Wynndow:
Yeah, that was exactly the reason. If in actuality it doesn't block, then may as well get rid.

job_definitions/release_pipeline.yml
```diff
@@ -34,18 +34,16 @@
...
+                  parameters: [
+                    [
+                      $class: 'ChoiceParameterDefinition',
+                      choices: 'Retry\nSkip',
+                      description: 'Failed functional tests action',
+                      name: 'action'
                     ]
-                }
+                  ]
                 echo "Action chosen was ${env.FAILED_TEST_ACTION}"
```
Wynndow: Not sure if intentional or not, but these `echo`s are now inside the script block that sets the env variable. They weren't before. I don't know what the effect of this is, maybe nothing, thought it worth checking though.

samuelhwilliams: 👍 fixed

samuelhwilliams:
:bump:

PR #88, opened 2018-05-25
Virus scan job

Wynndow:
Job to allow the very long running virus scanning script to be run from Jenkins instead of a dev's laptop. Also means we have a shared and slightly more permanent record of the outcome.

Nice.

### DIFF
--- a/job_definitions/general_docker_script_runner.yml
+++ b/job_definitions/general_docker_script_runner.yml
@@ -25,7 +25,7 @@
               USERNAME=General docker script runner
               JOB=See build for command run
               ICON=:railfail:
-              STATUS=FAILURE
+              STATUS=FAILED
               URL=<${BUILD_URL}consoleFull|#${BUILD_NUMBER}>
               CHANNEL=#dm-release
     builders:
--- a/job_definitions/export_framework_applicant_details.yml
+++ b/job_definitions/export_framework_applicant_details.yml
@@ -20,7 +20,7 @@
                 JOB=Export of framework applicant details
                 ICON=:user:
                 STAGE={{ environment }}
-                STATUS=PASSED
+                STATUS=SUCCESS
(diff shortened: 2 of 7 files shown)


job_definitions/general_docker_script_runner.yml
```diff
@@ -25,7 +25,7 @@
               USERNAME=General docker script runner
               JOB=See build for command run
               ICON=:railfail:
-              STATUS=FAILURE
+              STATUS=FAILED
```
samuelhwilliams: There are quite a few other jobs that set status/build result to `FAILURE`. We know this is probably wrong now, so might as well make them all consistent as part of this.

Wynndow: I've pushed an update so we use `FAILED` whenever we notify slack. That will mean the correct colour gets applied in the alert. I'm not sure that setting currentBuild.result changes anything specific to have left as is.

I've also added a lovely shade of blue we can use by setting the status for slack notifications to 'APPROVED'. Mostly for the visual regression tests. Let me know what you think.

samuelhwilliams: Yeah it's maybe safe(r) to leave the currentBuild.result things as they are, although I wonder if any of the slack jobs are sent with STATUS=currentBuild.result?

It feels like it would be nice to have consistent language (FAILED vs FAILURE), just to stop people seeing the wrong word, copying it and then not getting delightful colours (which is what happened to me I think).

Unless we can easily have both FAILED and FAILURE come up with colours? In that case I guess it doesn't matter _too_ much. But still feels nice to be consistent. 🤷‍♂️

Wynndow: I hit merge before reading this...

Wynndow: We could easily add something to the notify slack job. Anything that starts with an `F` or something

PR #95, opened 2018-06-20
Use docker for publishing services job

Wynndow:


samuelhwilliams:
Might be good to copy this over and make a job for DOS while we're at it.

PR #99, opened 2018-07-04
Fix nacl lib for nix shell

galund:
When installed via pip, the nacl lib was failing to compile.

* instead, install via nix. Seems to work :shrug:


risicle:
It's probably better to address the compilation error rather than end up with nix users having a different version of pynacl to everyone else... i'll message you...

risicle:
Just gonna check this WFM...

PR #104, opened 2018-07-17
stats_to_performance_platform: enable for dos3

risicle:


risicle:
See https://github.com/alphagov/digitalmarketplace-jenkins/pull/105

PR #105, opened 2018-07-17
stats_to_performance_platform: enable for dos3, change credentials format

risicle:
We now fetch the pp bearer token out of a per-`pp_service` dictionary, named `performance_platform_bearer_tokens`. Expect a PR to the credentials repo mirroring this soon.

We seem to be being provided with a new bearer token for each `pp_service` so we'll need to do this in case we're needed to have >1 framework open at once.


risicle:
Double approval!

PR #106, opened 2018-07-26
Remove API token arg from script

katstevens:
Trello: https://trello.com/c/XJEqu5gI/15-spike-audit-tokens-used-by-jenkins-jobs

Removes the token to align with the script changes (to avoid logging the credentials in Jenkins console).

Also deletes the ancient performance testing jobs that would also log credentials to Jenkins, if anyone ever tried to run them.

**Do not merge** before https://github.com/alphagov/digitalmarketplace-scripts/pull/265 is ready


risicle:
Have you actually applied this? The jenkins job was failing over the weekend

PR #108, opened 2018-07-27
Remove dry_run default = true from retention scripts after successful testing and production runs

benvand:
Remove dry_run default = true from retention scripts after successful testing and production runs


job_definitions/data_retention.yml
```diff
@@ -1,6 +1,6 @@
-{% set environments = [('preview', 'false'), ('staging', 'true'), ('production', 'true')] %}
+{% set environments = [('preview'), ('staging'), ('production')] %}
```
katstevens: Do these need to be tuples now?

lfdebrux: ➕

benvand: Done 👍

PR #109, opened 2018-07-27
Jenkins3

Wynndow:
This is the work that George did to update our ansible configuration to work with a new os and ansible version.

It's broadly the same as what he presented to us just before the end of the last mission, apart from a few commits at the end to tweak a couple of bits.

I have run all of this successfully onto a new box. You can currently see the results at https://ci3.marketplace.team


playbooks/roles/jenkins/tasks/plugins.yml
```diff
@@ -1,28 +1,121 @@
...
+
+- name: Check if we need to restart Jenkins to activate plugins
+  uri:
+    user: "{{ plugins_username }}"
+    password: "{{ plugins_password }}"
+    force_basic_auth: yes
+    url: "http://127.0.0.1:8080/updateCenter/api/json\
+    ?tree=restartRequiredForCompletion&{{ jenkins_crumb_token }}"
+    return_content: yes
+  register: jenkins_restart_required
```
katstevens: Could we put this in an Ansible `notify` handler for restarting Jenkins rather than doing the `wait` stuff above? Or does this have to be done before the next task?

playbooks/ansible.cfg
```diff
@@ -1,2 +1,3 @@
 [ssh_connection]
 control_path = %(directory)s/%%h-%%r
+ssh_args = -o ControlMaster=no
```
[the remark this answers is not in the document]
Wynndow: We added this a while back to fix a weird connection error, not 100% why we need it. ControlMaster if used, will allow multiple ssh connections to the same server to use the same tcp connection, which saves time when there's high latency. Turning it off prevented the ssh errors so we just went with it.

playbooks/roles/jenkins/tasks/plugins.yml
Wynndow: Yeah, this needs to be done before the next tasks are run when booting from scratch. If you just run all the tags in one go, the handlers only get run at the end of everything.

katstevens: Makes sense!

PR #114, opened 2018-08-16
Ensure latests Jenkins is installed

Wynndow:
We want to be able to bump the version to the latest to remain on top of
security issues.

### DIFF
--- a/playbooks/roles/jenkins/tasks/jenkins.yml
+++ b/playbooks/roles/jenkins/tasks/jenkins.yml
@@ -31,7 +31,9 @@
             backup=no
 
 - name: Ensure Jenkins is installed.
-  package: name=jenkins state=latest
+  apt:
+    name=jenkins=2.*
+    state=present
   register: jenkins_package_install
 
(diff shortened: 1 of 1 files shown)


playbooks/roles/jenkins/tasks/jenkins.yml
```diff
@@ -31,7 +31,7 @@
             backup=no
 
 - name: Ensure Jenkins is installed.
-  package: name=jenkins state=present
+  package: name=jenkins state=latest
```
katstevens: Can we stipulate a major version that `latest` can't go above? The Ansible docs aren't particularly helpful on this...

Wynndow: interesting....i'll have a look

Wynndow: It can be done!

katstevens: Excellent 👍

PR #123, opened 2018-08-28
Add `copy_old_builds.sh` to playbook

Wynndow:
If we're creating a new Jenkins box we're probably going to want to copy
the old builds from the outgoing box. There's a dead simple script to do
this using rsync. We can add it to the Jenkins build so it's available
without having to add it to the box later.

To use it, the ssh key being used will need filling in, as well the
hostname of the old server.

There didn't seem to be a great place to add this script copy, so I
created a new task. It needs to be run after the config task as this is
where the jenkins data directory is created with permissions.

### DIFF
--- a/playbooks/roles/jenkins/files/copy_old_builds.sh
+++ b/playbooks/roles/jenkins/files/copy_old_builds.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+for d in {release-*,functional-tests-*,visual-regression-*,build-image,database-migration-paas,tag-application-deloyment,clean-and-apply-db-dump-*,index-services-*,index-briefs-*,update-*-index-alias,data-retention-*,database-backup,rotate-api-tokens,rotate-production-notify-callback-token}/; do
+    rsync -vazh -e "ssh -i ~/.ssh/<private key to use>" --progress --include 'builds**' --include 'last*' --include 'nextBuildNumber' --exclude '*' jenkins@<source instance host>:/data/jenkins/jobs/"$d" /data/jenkins/jobs/"$d"
+done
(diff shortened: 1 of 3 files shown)


playbooks/roles/jenkins/files/copy_old_builds.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -o errexexit
```
katstevens: Should this be `errexit`?

playbooks/roles/jenkins/files/copy_old_builds.sh
```diff
@@ -0,0 +1,7 @@
...
+
+for d in {release-*,functional-tests-*,visual-regression-*,build-image,database-migration-paas,tag-application-deloyment,clean-and-apply-db-dump-*,index-services-*,index-briefs-*,update-*-index-alias,data-retention-*,database-backup,rotate-api-tokens,rotate-production-notify-callback-token}/; do
```
katstevens: We're missing a couple of jobs from this list (build-scripts, export-data, the smoke tests, a few others) - not sure if this list is meant to be exhaustive or if we're only transferring certain job histories?

playbooks/roles/jenkins/files/copy_old_builds.sh
Wynndow: Yes indeed!

playbooks/roles/jenkins/files/copy_old_builds.sh
Wynndow: No, it's only the jobs we really care about the history of. Release pipeline stuff and data backup stuff and a couple of others.

PR #126, opened 2018-09-04
New jobs to disambiguate the role of creating and updating indexes

benvand:
After chatting to @Wynndow  this morning, I think a lot of the difficulty in understanding what the various index jobs are for came from the fact that the `index_services` and `index_briefs` jobs did almost exactly the same thing. But crucially both behave very differently when you specify the `CREATE_WITH_MAPPING` param. In that case instead of updating an existing index they create a new one

There's quite a big cognitive distance between creating and updating an index. And the situations in which you'd want to perform those actions are also pretty different.

This PR has a couple of new jobs for Jenkins split along the lines of `create_index` and `update_index` rather than what objects they're operating on.

I'd add them in parallel then we can remove the usage of the old ones once these are tested.

### DIFF
--- a/job_definitions/create_index.yml
+++ b/job_definitions/create_index.yml
@@ -51,5 +51,9 @@
               CHANNEL=#dm-release
     builders:
       - shell: |
-          docker run -e DM_DATA_API_TOKEN_{{ environment|upper }} -e DM_SEARCH_API_TOKEN_{{ environment|upper }} digitalmarketplace/scripts scripts/index-to-search-service.py "${OBJECTS}" '{{ environment }}' --serial --index="${INDEX_NAME}" --frameworks="${FRAMEWORKS}" --create-with-mapping="${MAPPING}"
+          if [ "$SERIAL" = "true" ]; then
+            FLAGS="--serial"
+          fi
+
+          docker run -e DM_DATA_API_TOKEN_{{ environment|upper }} -e DM_SEARCH_API_TOKEN_{{ environment|upper }} digitalmarketplace/scripts scripts/index-to-search-service.py "${OBJECTS}" '{{ environment }}' --index="${INDEX_NAME}" --frameworks="${FRAMEWORKS}" --create-with-mapping="${MAPPING}" $FLAGS
 {% endfor %}
--- a/job_definitions/create_index.yml
+++ b/job_definitions/create_index.yml
@@ -1,0 +1,55 @@
+{% for environment in ['preview', 'staging', 'production'] %}
+- job:
+    name: "create-index-{{ environment }}"
+    display-name: "Create index - {{ environment }}"
+    project-type: freestyle
+    description: >
+      This job creates a new index in the ElasticSearch service from specified objects in the database.
+
+      Visiting either https://search-api.<STAGE>.marketplace.team/_status or
+      https://search-api.digitalmarketplace.service.gov.uk/_status will show you currently available indexes and their
+      aliases.
+
+      You can use the above to get an idea of naming conventions.
+
+      Once you have created your new index you might want to use it on the site. To do so, you can give your new index
+      an 'alias' here https://ci.marketplace.team/job/update-{{ environment }}-index-alias/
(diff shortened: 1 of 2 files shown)


job_definitions/create_index.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      Once you have created your new index you might want to use it on the site. To do so, you can give your new index
+      an 'alias' here https://ci.marketplace.team/job/update-{{ environment }}-index-alias/
+    parameters:
+      - choice:
+          name: OBJECTS
+          choices:
+            - briefs
+            - services
+          default: services
```
Wynndow: Why default to services?

job_definitions/create_index.yml
```diff
@@ -0,0 +1,56 @@
...
+      - string:
+          name: INDEX_NAME
+          default: {{ search_config['services'][environment].default_index }}
```
Wynndow: Would it be better to not supply a default at all? I'm guessing that most of the time these are going to be called by other Jenkins jobs, from where we can specify these details. Might help prevent an accident.

job_definitions/create_index.yml
```diff
@@ -0,0 +1,56 @@
...
+          description: "Comma-separated list of framework slugs that should be indexed (eg 'g-cloud-7,g-cloud-8'). If no frameworks are specified then all currently published services will be indexed."
+      - string:
+          name: MAPPING
+          description: >
+            This mapping is a filename (without .json suffix) which should exist in the
+            digitalmarketplace-search-api/mappings directory
+            (https://github.com/alphagov/digitalmarketplace-search-api/tree/master/mappings)
+            (eg 'briefs-digital-outcomes-and-specialists')
+    triggers:
+      - timed: "H 2 * * *"
```
Wynndow: Will this create a new index every morning? I think we just want to update indexes regularly overnight, not create new ones.

job_definitions/update_index.yml
```diff
@@ -0,0 +1,55 @@
...
+      be 'g-cloud-10'
+
+      However you may be testing a new index and want to input its name and frameworks manually.
+    parameters:
+      - choice:
+          name: OBJECTS
+          choices:
+            - briefs
+            - services
+          default: services
```
Wynndow: Again, not sure we should have defaults set.

job_definitions/update_index.yml
```diff
@@ -0,0 +1,55 @@
...
+          name: INDEX
+          default: {{ search_config['services'][environment].default_index }}
+          description: "The current name or alias of the index you want to update (briefs-digital-outcomes-and-specialists or g-cloud-10)"
+      - string:
+          name: FRAMEWORKS
+          description: >
+            Comma-separated list of framework slugs that should be indexed
+            (eg 'g-cloud-10' or 'digital-outcomes-and-specialists,digital-outcomes-and-specialists-2').
+    triggers:
+      - timed: "H 2 * * *"
```
Wynndow: As above.

job_definitions/create_index.yml
lfdebrux: Agreed.

job_definitions/update_index.yml
lfdebrux: This one should be timed right?

lfdebrux:
I'm not sure I understand the plan of adding them in parallel with the old scripts; will one or the other be disabled? Or will we be doing the updating twice each night?

job_definitions/update_index.yml
Wynndow: No, I don't think so. I see these two jobs as things that need to be wrapped and customised depending on if we're indexing briefs or services. I think the wrappers should be timed.

lfdebrux: Ah I think I understand now

benvand:
@lfdebrux I've implemented the serial flag the same way we implement `DRY_RUN` elsewhere.
Set as default = true because if people don't know what it does then they should be using it 🤔

PR #129, opened 2018-09-11
Job definition for new supplier CSV export script

katstevens:
Trello: https://trello.com/c/FtV7sVMU/138-admin-download-all-suppliers-for-framework-endpoint-locks-up-an-api-process

Jenkins job for the new supplier CSV script (see https://github.com/alphagov/digitalmarketplace-scripts/pull/277).

Ran for preview, the job and script (and error message to Slack) work as expected.

Not sure if hardcoding the framework slugs we want reports for is the best way to do this, but it'll represent what's in Jenkins (once I run this for staging and prod as well).


katstevens:
Thanks Chris - running this to create the staging/prod jobs now.

PR #132, opened 2018-09-19
Update description of "Apps are up" to accurately describe job

benvand:
Smoke tests now test 7 apps 

https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/features/smoke-tests/apps_status.feature
https://trello.com/c/uFfPSPZf/156-change-smoke-tests-description-on-jenkins

### DIFF
--- a/job_definitions/smoke_tests.yml
+++ b/job_definitions/smoke_tests.yml
@@ -6,13 +6,16 @@
     display-name: "Apps are up - {{ environment }}"
     project-type: pipeline
     description: |
-      Check that the five Digital Marketplace apps are up and running in the {{ environment|upper }} environment:
+      Check that the eight Digital Marketplace apps are up and running in the {{ environment|upper }} environment:
       <ul>
         <li>API</li>
         <li>Search API</li>
+        <li>Admin frontend</li>
+        <li>Briefs frontend</li>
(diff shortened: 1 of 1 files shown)


job_definitions/smoke_tests.yml
```diff
@@ -6,13 +6,16 @@
...
       <ul>
         <li>API</li>
         <li>Search API</li>
+        <li>Admin frontend</li>
+        <li>Briefs frontend</li>
+        <li>Brief-responses frontend</li>
         <li>Buyer frontend</li>
         <li>Supplier frontend</li>
-        <li>Admin frontend</li>
+        <li>User frontend</li>
```
katstevens: Should we include the router as an 'app'? The smoke tests certainly would fail without it 😸

benvand: My only problem with this is that if we mention it here we should really mention it on all the functional tests.

Saying that we could add the router to the smoke tests 🤔

benvand: Hmm maybe not though
I don't think we can get the url of a router app

katstevens: OK, makes sense.

PR #133, opened 2018-09-19
Add supplier CSV jobs to data tab

katstevens:
So we can find the new jobs easily.

I think this just needs the `make jenkins TAGS=config` to be run, but that will trigger a restart.

### DIFF
--- a/playbooks/roles/jenkins/defaults/main.yml
+++ b/playbooks/roles/jenkins/defaults/main.yml
@@ -83,6 +83,9 @@ jenkins_list_views:
       - database-backup
       - export-data-preview
       - export-data-production
+      - export-supplier-csv-preview
+      - export-supplier-csv-staging
+      - export-supplier-csv-production
       - index-services-preview
       - index-services-production
       - index-services-staging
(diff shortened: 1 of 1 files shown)


lfdebrux:
Your changes look good, but there is a fair amount of inconsistency in the list in general; some jobs don't have their staging version in the list (maybe they don't have a staging version?), others are not in preview staging production order... Do you mind if I tag in another commit to do a bit more clean up?

katstevens:
Go ahead 😸 Yep, some jobs don't exist for staging as we don't want to send 'real' emails.

playbooks/roles/jenkins/defaults/main.yml
```diff
@@ -107,18 +105,12 @@ jenkins_list_views:
...
-      - notify-buyers-to-award-closed-briefs-8-weeks-preview
       - notify-buyers-to-award-closed-briefs-8-weeks-production
-      - notify-buyers-when-requirements-close-preview
       - notify-buyers-when-requirements-close-production
-      - notify-suppliers-of-awarded-briefs-preview
       - notify-suppliers-of-awarded-briefs-production
       - notify-suppliers-of-dos-opportunities-production
       - notify-suppliers-of-new-questions-answers-production
-      - notify-suppliers-of-brief-withdrawals-preview
-      - notify-suppliers-of-brief-withdrawals-staging
```
katstevens: Oooh good spot on these...

katstevens:
Thanks Ben. Will run the config job on Jenkins before merging.

PR #134, opened 2018-09-25
Create a new pipeline job to release all the apps

benvand:
When inserting new credentials into the credentials repo we have to manually release all the apps, being _very_ careful not to miss anything, or use the rotate tokens job and manually release the router. Neither of these situations is very desirable and they are prone to human error. This job releases all the jobs on a given stage ensuring they all have a fresh version of the credentials.

Based on(/ a simplified version of):
https://github.com/alphagov/digitalmarketplace-jenkins/blob/master/job_definitions/rotate_api_tokens.yml#L77

### DIFF
--- a/job_definitions/rerelease_all.yml
+++ b/job_definitions/rerelease_all.yml
@@ -1,0 +1,73 @@
+{% set apps = ['router', 'search-api', 'api', 'antivirus-api', 'admin-frontend', 'briefs-frontend', 'brief-responses-frontend', 'buyer-frontend', 'supplier-frontend', 'user-frontend'] %}
+---
+- job:
+    name: rerelease-all-apps
+    display-name: "Re-release all apps"
+    project-type: pipeline
+    description: |
+        Update credentials and re-release all apps. Used for credentials changes to be brought in.
+
+        This performs a deploy of the listed apps using the `release_application_paas` job which determines their current version.
+        <br>See:<br>
+
+        https://docs.cloudfoundry.org/devguide/deploy-apps/start-restart-restage.html<br>
+        (Restart Your Application)<br><br>
+
+        https://github.com/alphagov/digitalmarketplace-jenkins/blob/master/job_definitions/release_application_paas.yml#L2<br>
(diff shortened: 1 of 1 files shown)


katstevens:
I can't remember - would we need to rebuild the scripts Docker image (i.e. run the `build-scripts` Jenkins job) to pull in new credentials there?

benvand:
@katstevens No they're not included in the docker image

benvand:
I've put a redundant step in to the job

To update credentials
I need feedback on whether this helps the job be more descriptive
Before they're implicitly updated by the app deployments
Now I explicitly run the job

job_definitions/rerelease_all.yml
```diff
@@ -0,0 +1,73 @@
+{% set apps = ['router', 'search-api', 'api', 'antivirus-api', 'admin-frontend', 'briefs-frontend', 'brief-responses-frontend', 'buyer-frontend', 'supplier-frontend', 'user-frontend'] %}
+---
+- job:
+    name: rerelease-all-apps
+    display-name: "Re-release all apps"
```
katstevens: Can we call this `Re-release all apps with latest credentials` so it's super clear what's happening?

job_definitions/rerelease_all.yml
```diff
@@ -0,0 +1,73 @@
...
+            string(name: 'CHANNEL', value: "#dm-release"),
+            text(name: 'STAGE', value: "${STAGE}"),
+            text(name: 'STATUS', value: status),
+            text(name: 'URL', value: "<${BUILD_URL}consoleFull|${BUILD_DISPLAY_NAME}>")
+          ]
+        }
+        currentBuild.displayName = "#${BUILD_NUMBER} - ${STAGE}"
+        notify_slack(':spinner:', 'Start re-release all apps on ${STAGE}', 'STARTED')
+        try {
+          stage('Update credentials') {
```
katstevens: Nitpicking again with naming, but could we call this `Pull in latest credentials` instead of 'update' - at first I thought this step was actually creating new credentials.

PR #145, opened 2018-10-04
Remove confusing 'finished' message on slack job

katstevens:
Trello: https://trello.com/c/GTueLBfq/181-notify-slack-job-adds-finished-to-every-message

Possibly the quickest Quick Win that ever won?


katstevens:
Applied!

PR #148, opened 2018-10-11
Rewrite description for update-index-alias job

lfdebrux:
When I was trying to fix the preview smoke tests I needed to use the update-index-alias job, but I found the description a little confusing, so I've tweaked it here.

Feel free to bikeshed, the main thing I care about is that the example for TARGET actually looks like one of our index names, rather than one of our alias names.

### DIFF
--- a/job_definitions/update_index_alias.yml
+++ b/job_definitions/update_index_alias.yml
@@ -2,26 +2,30 @@
 - job:
     name: "update-{{ environment }}-index-alias"
     display-name: "Update {{ environment }} index alias"
-    description: Apply a given alias to a given elasticsearch index
+    description: >
+        Create an alias for a given elasticsearch index.
+        If there is an index with that alias already then
+        that index will be given the alias '<alias>-old'.
     project-type: pipeline
     concurrent: false
     parameters:
       - string:
           name: ALIAS
           default:
-          description: "The name of the alias you're applying."
+          description: "The alias name."
       - string:
           name: TARGET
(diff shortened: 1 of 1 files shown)


job_definitions/update_index_alias.yml
```diff
@@ -2,26 +2,30 @@
 - job:
     name: "update-{{ environment }}-index-alias"
     display-name: "Update {{ environment }} index alias"
-    description: Apply a given alias to a given elasticsearch index
+    description: >
+        Create an alias for a given elasticsearch index.
+        If there is an index with that alias already then
+        then index will be given the alias '<alias>-old'.
```
risicle: > then then

lfdebrux: 🤦‍♀️

job_definitions/update_index_alias.yml
```diff
@@ -2,26 +2,30 @@
...
     project-type: pipeline
     concurrent: false
     parameters:
       - string:
           name: ALIAS
           default:
-          description: "The name of the alias you're applying."
+          description: "The alias name."
```
katstevens: Let's have an example here too `e.g. 'g-cloud-10'`

job_definitions/update_index_alias.yml
katstevens: We could add a sentence here saying what we actually did: `Look at previous builds for example parameters.`

PR #150, opened 2018-10-22
virus-scan-s3-bucket job: upgrade to use api-calling script 

risicle:
https://trello.com/c/eI2PEVIY/202-create-jenkins-job-for-virus-scanning-catchup-script

Also includes an upgrade of the `pollscm` syntax usage to please new JJB.

Not included in this: periodic triggers, but I'm presenting this now because it's what's already been pushed to jenkins.

### DIFF
--- a/job_definitions/virus_scan_s3_bucket.yml
+++ b/job_definitions/virus_scan_s3_bucket.yml
@@ -38,4 +38,31 @@
           fi
 
           docker run -e DM_ANTIVIRUS_API_TOKEN_{{ environment|upper }} digitalmarketplace/scripts scripts/virus-scan-s3-bucket.py {{ environment }} ${BUCKETS} --concurrency="${CONCURRENCY}" --prefix="${PREFIX}" ${FLAGS}
+- job:
+    name: "virus-scan-s3-buckets-nightly-{{ environment }}"
+    display-name: "Virus scan S3 buckets nightly - {{ environment }}"
+    project-type: freestyle
+    triggers:
+      - timed: "H 3 * * *"
+    publishers:
+      - trigger-parameterized-builds:
+          - project: notify-slack
+            condition: UNSTABLE_OR_WORSE
+            predefined-parameters: |
+              USERNAME=virus-scan
+              JOB=Virus scan S3 buckets nightly - {{ environment }}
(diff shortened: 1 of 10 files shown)


job_definitions/virus_scan_s3_bucket.yml
```diff
@@ -1,49 +1,27 @@
...
-              CHANNEL=#dm-release
-          - project: notify-slack
-            condition: UNSTABLE_OR_WORSE
-            predefined-parameters: |
-              USERNAME=Jenkins
-              JOB=Virus scan S3 bucket
-              ICON=:can-of-worms:
-              STATUS=FAILED
-              URL=<${BUILD_URL}consoleFull|#${BUILD_NUMBER}>
-              CHANNEL=#dm-release
```
katstevens: I know we send emails if the AV API finds something bad, but what if the script fails for another reason - bad API key or someone's been fiddling with the script repo. (Agree that we probably don't need the success trigger.)

risicle: Um, yup, sure, we should probably alert on this.

 - [ ]

risicle: (or maybe the nightly jobs themselves?)

risicle:
How's everyone feeling about this now?

playbooks/roles/jenkins/defaults/main.yml
```diff
@@ -126,7 +126,12 @@ jenkins_list_views:
       - update-credentials
       - rotate-api-tokens
       - rotate-production-notify-callback-token
-      - virus_scan_s3_bucket
+      - virus-scan-s3-buckets-preview
+      - virus-scan-s3-buckets-staging
+      - virus-scan-s3-buckets-production
+      - virus-scan-s3-buckets-nightly-preview
+      - virus-scan-s3-buckets-nightly-staging
+      - virus-scan-s3-buckets-nightly-production
```
katstevens: So we have nightly jobs - when would we use the other ones?

risicle: I'm using the other ones right now. When we want to run them with custom parameters basically. Big scan jobs etc.

PR #151, opened 2018-10-24
487 smoke test silent failures

benvand:
* Create a new 'Prepare' step for smoke tests to bring in line with other pipelines
* Extend try/ catch to capture git checkout

### DIFF
--- a/job_definitions/smoke_tests.yml
+++ b/job_definitions/smoke_tests.yml
@@ -39,15 +39,15 @@
                   text(name: 'URL', value: "<${BUILD_URL}smoke_test_report|${BUILD_DISPLAY_NAME}>")
                 ]
         }
-
-        node {
-          git url: 'git@github.com:alphagov/digitalmarketplace-functional-tests.git', branch: 'master', credentialsId: 'github_com_and_enterprise'
-          stage('Run smoke tests') {
-            try {
+        try {
+          node {
+            stage('Prepare') {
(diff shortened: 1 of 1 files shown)


job_definitions/smoke_tests.yml
```diff
@@ -39,15 +39,15 @@
...
         }
-
-        node {
-          git url: 'git@github.com:alphagov/digitalmarketplace-functional-tests.git', branch: 'master', credentialsId: 'github_com_and_enterprise'
-          stage('Run smoke tests') {
-            try {
+        try {
+          node {
+            stage('Prepare') {
+              git url: 'git@github.com:alphagov/digitalmarketplace-functional-tests.git', branch: 'master', credentialsId: 'github_com_and_enterprise'
```
katstevens: Are we going to include any `wipe-workspace` flags or repository cleaning (I see in a couple of other jobs we do combos of `git clean -fdx` and `git reset --hard origin/master`)?

benvand: It feels a bit like treating the symptom rather than the cause. We shouldn't _have_ to include it. Lets chat in the 10 o'clock.

katstevens: I double checked and I think we only do the git clean etc for the `generate_upload_and_notify_counterpart_signature-pages` job (which is a bit of a unique case, it's the only regular script of ours not run in a Docker image). Let's try without it.

benvand: Is this good for an approve?

katstevens: Yep looks good.

PR #152, opened 2018-10-24
Separate playbook for upgrading packages

katstevens:
Trello: https://trello.com/c/OujrG25u/486-update-git-version-on-jenkins

This playbook would be run with `make jenkins TAGS=upgrade_packages`.

For DRYness I've put the package list into the defaults file.

I haven't run it yet! Locally I have `check_mode: yes` enabled, I can try it out tomorrow unless anyone spots something weird.

### DIFF
--- a/playbooks/roles/jenkins/defaults/main.yml
+++ b/playbooks/roles/jenkins/defaults/main.yml
@@ -16,6 +16,30 @@ jenkins_java_args: ""
 jobs: '*'
 jobs_disabled: false
 
+dist_tools:
+  - git
+  - libpq-dev
+  - python-dev
+  - libyaml-dev
+  - libreadline-dev
+  - libffi-dev
+  - postgresql-client-9.5
+  - curl
+  - zip
+  - unzip
+  - ntp
+  - jq
+  - xvfb
+  - wkhtmltopdf
+  - fonts-liberation
+  - gnupg2
+  - make
+  - zlib1g-dev
+  - bzip2
+  - libsqlite3-dev
+  - libssl-dev
+  - python3-pip
+
 dm_applications:
   - api
   - search-api
--- a/playbooks/roles/jenkins/tasks/upgrade_packages.yml
+++ b/playbooks/roles/jenkins/tasks/upgrade_packages.yml
@@ -1,0 +1,6 @@
+---
+# Upgrades packages separately from the main Jenkins build playbook
+- name: Upgrade all packages to the latest version
+  tags: [apt]
+  apt: update_cache=yes name={{ item }} state=latest
+  with_items: "{{ dist_tools }}"
--- a/playbooks/roles/jenkins/tasks/tools.yml
+++ b/playbooks/roles/jenkins/tasks/tools.yml
@@ -24,29 +24,7 @@
 - name: Install tools
   tags: [apt]
   apt: update_cache=yes name={{ item }} state=present
-  with_items:
-    - git
-    - libpq-dev
-    - python-dev
-    - libyaml-dev
-    - libreadline-dev
-    - libffi-dev
-    - postgresql-client-9.5
-    - curl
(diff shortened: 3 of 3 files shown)


lfdebrux:
This is pretty cool, but I'm surprised there wasn't an `apt-get ugrade` option?

playbooks/roles/jenkins/defaults/main.yml
```diff
@@ -16,6 +16,30 @@ jenkins_java_args: ""
...
+  - libreadline-dev
+  - libffi-dev
+  - postgresql-client-9.5
+  - curl
+  - zip
+  - unzip
+  - ntp
+  - jq
+  - xvfb
+  - wkhtmltopdf
```
samuelhwilliams: I think there was an issue at some point where an upgrade to this caused PDFs to be rendered in a much smaller font size. I don't know the exact context (I think Chris handled it) - I just overhead it. Just something to be aware of/mitigate against if the team is planning to bump these libs frequently - there's probably a reasonable risk of regressions somewhere in here. `jq` feels like another lib that _could_ change and break things.

playbooks/roles/jenkins/tasks/upgrade_packages.yml
```diff
@@ -0,0 +1,6 @@
+---
+# Upgrades packages separately from the main Jenkins build playbook
+- name: Upgrade all packages to the latest version
```
lfdebrux: Might be worth rewriting to this to explicitly state that it is Jenkins requirements, rather than all installed packages?

playbooks/roles/jenkins/defaults/main.yml
katstevens: Can't find that conversation, but I've done some digging around where these libraries are used:
- `wkhtmltopdf`: the `generate-framework-agreement-counterpart-signature-pages.py` scripts are not run in Docker images, so for now this still needs to be installed in Jenkins
- `jq`: only used in the two scripts where we generate pull requests (`rotate-api-tokens.sh` and `maintenance-mode-pr.sh` ), which are both run in Docker. `jq` is already a dependency in that Dockerfile.

I think the rest we can probably trust ubuntu to update? What do people think?

katstevens: Installing `wkhtmltopdf` in a separate step might be wise, too - like we do with `awscli`.

lfdebrux: It would be good to understand if the rest of the packages are installed because Jenkins needs them, or because we need them for our jobs... that might be a fair amount of work though.

katstevens: I've done a (not particularly thorough) search and apart from obvious ones like `make`, `git` and `curl`, I can't see direct usage in the jenkins jobs/scripts. However I think George and Chris were pretty thorough when revamping all these ansible playbooks, so I'm sure these are all needed *somewhere* 😸 I guess for now we can pin the ones we know need pinning for certain, let the others upgrade, see if it causes problems? `:this_is_fine_dog:`

katstevens:
I've gone through the whole list of dist tools and the versions we have on the box currently match the xenial LTS package versions, except for the following: `python-dev`, `zlib1g-dev` (which match the xenial updates versions) and `libpq-dev` which is actually a major bump ahead.

playbooks/roles/jenkins/tasks/upgrade_packages.yml
katstevens: Well, everything on the box is in theory a Jenkins requirement 😸 These are ubuntu packages so I'd rather keep it generic.

katstevens: I will however change the comment to make it clear we're sticking to LTS versions.

PR #154, opened 2018-10-28
nightly virus-scan-s3-buckets job: only notify slack for production

risicle:
We don't want to be woken by failures on preview.

### DIFF
--- a/job_definitions/virus_scan_s3_bucket.yml
+++ b/job_definitions/virus_scan_s3_bucket.yml
@@ -44,6 +44,16 @@
     project-type: freestyle
     triggers:
       - timed: "H 3 * * *"
+    builders:
+      - trigger-builds:
+          - project: "virus-scan-s3-buckets-{{ environment }}"
+            condition: UNSTABLE_OR_WORSE
+            block: true
+            predefined-parameters: |
+              BUCKETS=digitalmarketplace-agreements-{{ environment }}-{{ environment }},digitalmarketplace-communications-{{ environment }}-{{ environment }},digitalmarketplace-documents-{{ environment }}-{{ environment }},digitalmarketplace-submissions-{{ environment }}-{{ environment }}
+              SINCE=49 hours ago
+              DRY_RUN=false
+{% if environment != 'preview' %}
     publishers:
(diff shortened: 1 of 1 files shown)


lfdebrux:
Are these guaranteed to fail on preview because of EICAR files?

risicle:
No - those should get scanned and tagged thru their SNS notification. An already-tagged file will not be rescanned and not raise an alert.

lfdebrux:
Hmm, so we should expect the tests to not fail most of the time?

risicle:
The tests assert the behaviour of the SNS loop - they look for the (dummy) email sent as a result of the SNS message being processed. The catchup script failed on friday night because i'd been dicking with the av-api on friday making it miss certain SNS messages, leaving some eicar-ed files untagged.

lfdebrux:
If we're not expecting the tests to fail often then I'm happy for the notifications to remain; I think it is up to devs to make sure they have their phones off during off-hours 😜

risicle:
My point is that this is for preview, where the primary cause of "failure" will be due to people dicking around with it during the previous day. Not something that's a second line issue.

I'm half tempted to remove the slack alert entirely, because found viruses result in an email going to the mailing list anyway, whether triggered by SNS or the script.

lfdebrux:
I see your point.

It should probably be only for preview though; currently this PR also changes staging as I understand it?

katstevens:
> I'm half tempted to remove the slack alert entirely, because found viruses result in an email going to the mailing list anyway, whether triggered by SNS or the script.

I think I said this before, but we won't get an email if there's a failure executing the script itself (or sending the email, or git failing to checkout the repo properly). So keep the slack alert 😸

risicle:
Uh yeah ok.

risicle:
@katstevens indeed, but we don't have slack alerts for *all* our nightly jobs do we?

katstevens:
We do for pretty much all our email scripts, indexing and data export jobs (for `condition: UNSTABLE_OR_WORSE`).

risicle:
Oh ok fine. Keeping for staging & prod.

PR #155, opened 2018-10-30
Fix update index alias job

lfdebrux:
Recently this job failed because the pip requirements file for digitalmarketplace-scripts was in a bad state. However, this should not have caused the job to fail, because this script was being run in a Docker container.

This PR removes the dead code that caused the failure.

It also renames the update index alias job naming to be consistent with other similar jobs.

### DIFF
--- a/job_definitions/clean_and_apply_db_dump.yml
+++ b/job_definitions/clean_and_apply_db_dump.yml
@@ -111,7 +111,7 @@
                   ]
                 }
                 stage('Update index alias'){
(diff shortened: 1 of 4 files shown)


lfdebrux:
So originally I changed the job from a pipeline to a freestyle job, but that messed up the ability to preserve history, so I'm keeping this as a pipeline job now

lfdebrux:
This config is now up on Jenkins

PR #160, opened 2018-11-01
Add local tests for job templates

lfdebrux:
Currently the only way to test job template files is by running the full Ansible playbook.

This is undesirable as it is a slow way to check for simple syntax errors, and it is dangerous as we only have one Jenkins box that deals with everything in production.

This is an especially pronounced problem when we are going to try a [major upgrade of Jenkins Job Builder](https://trello.com/c/pSEUci7F/), which may require substantial changes to our job templates.

---

The method used in this PR to test our job definition files tries to stick as closely as possible to that used to generate job files for our Jenkins server; it uses a custom Ansible playbook to generate the yaml files suitable for jenkins-jobs, and the version of jenkins-jobs and associated tools has been pulled out into a separate requirements file that used both for testing and on our Jenkins box.

This method does have the downside that it will require some maintenance if our playbooks chage, however I think the upsides are worth the extra moving parts.


lfdebrux:
Closed and reopened to trigger our *new* Travis job

lfdebrux:
For some reason the requirements didn't install properly; I might see whether rewriting this Makefile to be more like the ones in the other repos helps.

lfdebrux:
:tada:

test/fixtures/vars.yml
```diff
@@ -0,0 +1,106 @@
...
+    ci.marketplace.team:
+        client_id: __jenkins_github_auth_by_hostname____team____client_id__
+        client_secret: __jenkins_github_auth_by_hostname____team____client_secret__
+jenkins_github_email: __jenkins_github_email__
+java_certificate_store_pass: __java_certificate_store_pass__
+google_username: __google_username__
+google_password: __google_password__
+jenkins_gdrive_refresh_token: __jenkins_gdrive_refresh_token__
+jenkins_gdrive_db_dumps_folder_id: __jenkins_gdrive_db_dumps_folder_id__
+jenkins_gdrive_csv_data_export_folder_id: __jenkins_gdrive_csv_data_export_folder_id__
```
risicle: What were the "tweaks by hand" & were they significant? I ask because it would be nice if it could generate this on the fly so we don't problems with divergence...

lfdebrux: They were fairly significant. I'm working on it, but I didn't want to delay this PR any longer.

README.md
```diff
@@ -74,6 +74,16 @@ To upgrade a plugin (for example, to address a security vulnerability), tick the
...
 
+## Jobs
+
+Our Jenkins jobs are generated on the Jenkins box using the `jenkins-job-builder` tool via Ansible. If you want to test your job file locally before uploading, you can use
+
+```bash
+$ make test [JOBS=<job_name>]
+```
+
+If you want to test all of the job files it can be a bit slow, which is why `make test` will try and pre-cache the output from Ansible the first time you run it.
```
risicle: This is slightly unclear - will a regular `make test` test all jobs then?

lfdebrux: Yes, it will.

What I was trying to document is that if you run `make test JOBS=<job_name>` and you haven't run it before it will copy all of the Ansible files for you, which might be slightly surprising behaviour.

It does this because if you run `make -C test` without a cache, ansible will run once for each file, which takes a long time.

lfdebrux: It's not great behaviour, but again it can be fixed later.

test/fixtures/vars.yml
lfdebrux: Python script is now included

README.md
risicle: Ohhhhhkay.

lfdebrux: The Makefile is kind of a mess tbh, if you want we can pair on it.

test/Makefile
```diff
@@ -0,0 +1,78 @@
...
+	@tput setaf 6
+	@echo $1
+	@tput sgr0
+endef
+
+define capture-output
+	script $1 $2 > /dev/null || cat $1
+endef
+
+ROOT ?= ..
```
benvand: I think the `ROOT` var here could use a more descriptive name, PACKAGE_ROOT / DM_JENKINS_ROOT maybe?

test/Makefile
```diff
@@ -0,0 +1,78 @@
...
+VIRTUALENV_ROOT := $(ROOT)/venv
+PATH := $(VIRTUALENV_ROOT)/bin:$(PATH)
```
benvand: Same with PATH, VIRTUALENV_BINARY?
Also probably don't need to define VIRTUALENV_ROOT for a single use

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: python
+python:
+  - "2.7"
+install:
+  - make requirements-test
+  - make -C test pre-populate-cache
+script:
+  - make -C test
```
benvand: So this line just runs everything on the base/ root level of the test/ makefile?

test/Makefile
```diff
@@ -0,0 +1,78 @@
+comma := ,
```
benvand: convention elsewhere is to call the test dir `tests` by the looks of it

test/Makefile
```diff
@@ -0,0 +1,78 @@
...
+
+define process-templates-quiet
+	$(call capture-output,$(@:.yml=).log,$(call process-templates,$1,$2))
+endef
+
+$(ANSIBLE_OUTDIR):
+	@mkdir -p $@
+	@$(call colorecho,'Detected first run; generating cache...')
+	@$(call process-templates,$@,$(SOURCES))
+	@$(eval ALREADY_RAN := true)
```
benvand: Does it take a lot longer to rebuild the cache for each test run? Not sure we want it handing around 🤷‍♂️

playbooks/roles/jenkins/tasks/jobs.yml
```diff
@@ -5,10 +5,7 @@
 
 - name: copy a requirements file with pinned Python dependencies
   copy:
-    content: |
-      python-jenkins==0.4.15  # newer ones have broken basic auth :(
-      jenkins-job-builder==1.6.2
-      git+https://github.com/rusty-dev/jenkins-job-builder-pipeline.git@f24b9f5b0de03edd59b94061fa5182d11887403d#egg=jenkins_job_builder_pipeline
+    src: ../../../../requirements-jenkins-job-builder.txt
```
benvand: Is there a way we can just consolidate these with the ansible depoendency in requirements.txt and install them all for tests?

Ansible is a requirement of this package but we don't test it anywhere. That doesn't mean it shouldn't be installed when we're running tests. It could have conflicting dependencies. with one of the packages here.

test/Makefile
lfdebrux: I timed it as being 2 minutes instead of 30 seconds, which I thought was a pain, which is why I added the caching. But Make magically deletes the intermediate files at the end if you run `make test` without generating the cache first so it doesn't take up too much disk space.

test/Makefile
lfdebrux: Will rename

.travis.yml
lfdebrux: It runs the default goal which is currently `test-job-templates`

test/Makefile
lfdebrux: Fair

lfdebrux: PATH is the Unix PATH, can't change the name for that

lfdebrux: VIRTUALENV_ROOT is the same as in the root Makefile so don't really want to change that either...

test/Makefile
lfdebrux: Fair

lfdebrux: Renamed to `DM_JENKINS_REPO`

.travis.yml
lfdebrux: I've changed this line to be `make test` which is more self-descriptive

tests/Makefile
```diff
@@ -0,0 +1,79 @@
...
+pre-populate-cache: $(ANSIBLE_OUTDIR)
+
+define process-templates
+	ansible-playbook \
+		--inventory="localhost," \
+		--extra-vars=@$(DM_JENKINS_REPO)/playbooks/roles/jenkins/defaults/main.yml \
+		--extra-vars=@$(DM_JENKINS_REPO)/tests/fixtures/vars.yml \
+		--extra-vars="dest=$1" \
+		--extra-vars="items=[$(subst $(space),$(comma),$2)]" \
+		$(DM_JENKINS_REPO)/tests/test_templates.yml
```
katstevens: I think we should state more clearly the scope of what we're testing. It doesn't pick up bad yaml, for example (changing `concurrent: true` to `concurrent: bananas` was ignored) or any syntax problems within node/shell `script`. That's fine, but we should be clear to the user what a green test implies.

Also, I ran this once and got a bunch of warnings, but re-running gives nothing - is there an option where I can clear that cache?

tests/fixtures/_gen_stubs.py
```diff
@@ -0,0 +1,25 @@
...
+        ]
+    else:
+        return prefix
+
+if __name__ == "__main__":
+    import yaml
+    import sys
+
+    yaml.dump(
+        tree_paths(yaml.load(sys.stdin)),
```
katstevens: Can we double check that we're using the version of `yaml` with safe load - I can't remember if the fix made it here already.

tests/Makefile
lfdebrux: > I think we should state more clearly the scope of what we're testing.

Good point. That might also encourage further tests to be added in future!

> (changing concurrent: true to concurrent: bananas was ignored)

That's a bit concerning, although both are valid YAML I would expect only one of those to be valid for Jenkins.

> Also, I ran this once and got a bunch of warnings, but re-running gives nothing - is there an option where I can clear that cache?

I ran into the same thing, I wasn't sure how to deal with it. It would be nice for the output to be more like a linter, in that it always checks everything, but then it's kind of nice to have the ability to run `make test` and have only the changed files checked.

lfdebrux: I guess if we wanted it to act more like a linter then the complicated Makefile wouldn't be needed, it could be a more straightforward script... That's an option I hadn't considered before.

Opinions more than welcome.

tests/fixtures/_gen_stubs.py
lfdebrux: Good question!

lfdebrux: The latest version of PyYAML is pulled in from somewhere, so we could use `safe_load` if we wanted to retain this script.

tests/Makefile
```diff
@@ -0,0 +1,79 @@
...
+	@$(if $(ALREADY_RAN),,$(call process-templates-quiet,$(@D),$?))
+
+$(JJB_OUTDIR)/%.xml: $(ANSIBLE_OUTDIR)/%.yml
+	@mkdir -p $(@D)
+	@$(call colorecho,'Parsing `$(<F)` with jenkins-jobs...')
+	@tput setaf 1 # print jenkins-jobs warnings in red
+	@jenkins-jobs -l=WARNING test $< > $@
+	@tput sgr0
+
+generate-fixtures:
```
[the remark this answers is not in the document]
lfdebrux: Votes here @benvand @katstevens @risicle, 👍 for keep, 👎 for not keep.

tests/Makefile
lfdebrux: > > (changing concurrent: true to concurrent: bananas was ignored)
> 
> That's a bit concerning, although both are valid YAML I would expect only one of those to be valid for Jenkins.

Might be worth working [jenkins-job-linter](https://github.com/OddBloke/jenkins-job-linter) into the tests as well.

lfdebrux:
Thanks everyone for reviewing this PR, I'm going to have another crack at how to do this, I think it's clear from the discussion here that I've come at this from slightly the wrong angle, but your feedback has helped me understand what we want better.

I'm going to close this PR to preserve the branch and discussion, and also to clear the board for take 2.

tests/Makefile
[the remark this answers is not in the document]
katstevens: Leaning towards 👎 -  I think the usefulness would be outweighed by the cost.

tests/Makefile
katstevens: 👍 for separate linting - ideally as a first step in `make test` (i.e. following our other linting conventions)

risicle:
This hasn't gone for good has it? Not having it sucks...

PR #161, opened 2018-11-02
Fix logrotate warnings

lfdebrux:
logrotate is deprecated on jenkins>=1.637, use the property build-discarder on newer jenkins instead


katstevens:
Does this need a specific version of JJB? Could we merge this without https://github.com/alphagov/digitalmarketplace-jenkins/pull/160?

lfdebrux:
The error is from the version of JJB we currently use. I caught the warning using the tool in #160 but the warning is emitted wherever/however you run JJB. There's no reason not to merge before #160 that I can think of.

PR #162, opened 2018-11-06
Add Jenkins Job Builder lint tool

lfdebrux:
This is a second attempt at [adding local tests for our Jenkins Job Builder templates](https://github.com/alphagov/digitalmarketplace-jenkins/pull/160), which I think we need so that we can easily [update Jenkins Job Builder](https://trello.com/c/pSEUci7F/).

It is a bit simpler than #160 in the sense that it uses a bash script instead of a Makefile, but it is still pretty complicated. I looked into creating a Python tool but it would have involved having to use a bunch of private APIs for both ansible and JJB.

Comments and opinions are always welcome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: python
+python:
+  - "2.7"
+install:
+  - sudo apt-get install jq
+  - make requirements-test
+script:
+  - make test ARGS=-v
--- a/Makefile
+++ b/Makefile
@@ -14,15 +14,19 @@ endif
 help: ## List available commands
 	@cat $(MAKEFILE_LIST) | grep -E '^[a-zA-Z_-]+:.*?## .*$$' | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
-.PHONY: requirements
-requirements: venv ## Install requirements
-	${VIRTUALENV_ROOT}/bin/pip install -Ur requirements.txt
-
 .PHONY: venv
 venv: ${VIRTUALENV_ROOT}/activate ## Create virtualenv if it does not exist
 
 ${VIRTUALENV_ROOT}/activate:
 	[ -z $$VIRTUAL_ENV ] && [ ! -d venv ] && python3 -m venv venv || true
+
+.PHONY: requirements
+requirements: venv ## Install requirements
+	${VIRTUALENV_ROOT}/bin/pip install -Ur requirements.txt
+
+.PHONY: requirements-test
+requirements-test: requirements ## Install test requirements
+	${VIRTUALENV_ROOT}/bin/pip install -Ur requirements-jenkins-job-builder.txt
 
 .PHONY: clean
 clean: ## Clean workspace (delete all generated files)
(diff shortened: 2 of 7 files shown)


lfdebrux:
Travis CI tests are running as expected, unfortunately they are failing due to warnings being issued by `jenkins-jobs test` :sweat_smile:

Makefile
```diff
@@ -14,16 +14,27 @@ endif
...
+.PHONY: requirements-test
+requirements-test: requirements ## Install test requirements
+	${VIRTUALENV_ROOT}/bin/pip install -Ur requirements-jenkins-job-builder.txt
+ifeq (, $(shell which jq))
+	$(error jjb-lint.sh requires jq, please ensure it is installed)
+endif
+
+.PHONY: test
+test: requirements-test
+	source ${VIRTUALENV_ROOT}/bin/activate && tools/jjb-lint/jjb-lint.sh job_definitions/*.yml
```
benvand: Dunno if you can use the existing venv command on l17 here?

requirements-jenkins-job-builder.txt
```diff
@@ -0,0 +1,3 @@
+python-jenkins==0.4.15  # newer ones have broken basic auth :(
```
benvand: we can't add these to `requirements.txt` can we? I'm pretty sure I figured out there was a good reason in the last PR if you can jog my memory 😁

lfdebrux: I think the reason was because we want to use `requirements-jenkins-job-builder.txt` on the Jenkins box, and if we add it to `requirements.txt` then we will end up installing Ansible on the Jenkins box. That might not be an issue though.

Makefile
lfdebrux: I think l17 installs the virtualenv but doesn't activate it.

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: python
+python:
+  - "2.7"
+install:
+  - sudo apt-get install jq
```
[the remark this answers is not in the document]
lfdebrux: Seems to work :man_shrugging:

lfdebrux:
Okay I think I'm done here.

Once #163 is merged I will rebase this on master, squash most of the commits to make a cleaner history, and then merge provided no one has any objections.

tools/jjb-lint/jjb-lint.sh
```diff
@@ -0,0 +1,376 @@
...
+}
+
+copy_jjb_config()
+{
+	# We want to test jenkins-jobs with the
+	# configuration used in production, but
+	# we want to tweak it slightly to avoid
+	# jenkins-jobs contacting jenkins
+
+	run_ansible "jenkins_jobs.ini.out" "${DM_JENKINS_REPO}/playbooks/roles/jenkins/templates/jenkins_jobs.ini.j2" > /dev/null
```
katstevens: Which bits of the config exactly are we tweaking? Presumably the user/password?

lfdebrux: Ah, that could definitely be better commented.

I will add this comment:

```bash
# If jenkins-jobs finds a configuration file it tries to contact the Jenkins server
# to detect what plugins are installed, unless `query_plugins_info` is set to false
# in the configuration file.
# See <https://docs.openstack.org/infra/jenkins-job-builder/execution.html#jenkins-section>
```

lfdebrux:
Tests passed because I've rebased on master (which has fixes for job definition issues)

lfdebrux:
Closing because the latest version of `jenkins-jobs` breaks this 😭 

This PR will return in Part III.

PR #164, opened 2018-11-14
Update Jenkins Job Builder to 2.6.0

lfdebrux:
This PR updates Jenkins Job Builder to the latest stable version, closing the outstanding [Trello ticket].

Only small changes were required to the job definition files (changing the `pipeline: script` key to `dsl`), the main blocker was the interaction between jenkins-job-builder and jenkins-job-builder-pipeline. It turns out that having jenkins-job-builder-pipeline with later versions of jenkins-job-builder will cause pipeline job definitions to fail. This was fixed by adding a task to the Ansible playbook to ensure that jenkins-job-builder-pipeline is uninstalled from the system Python environment.

These changes have now been manually tested on the Jenkins box.

[Trello ticket]: https://trello.com/c/pSEUci7F/103-upgrade-jenkins-job-builder-to-220


lfdebrux:
> Probs not worth the force uninstall of jjb-pipeline but otherwise looks awesome! Nice one 👍

Removed.

PR #169, opened 2018-11-21
smoke tests: make these also able to generate jenkins jobs for smoulder tests

risicle:
Is this an approach that people think is worth pursuing? I obviously haven't been able to try it yet...

The next thing I need is for the jenkins `smoke_test_vars` to be (initially) cloned to `smoulder_test_vars` in the credentials


job_definitions/smoke_tests.yml
```diff
@@ -1,42 +1,86 @@
...
-                string(name: 'USERNAME', value: 'smoke-tests'),
+                string(name: 'USERNAME', value: '{{ instance.slack_username }}'),
                 string(name: 'ICON', value: icon),
-                string(name: 'JOB', value: "Apps are up - {{ environment }}"),
+                string(name: 'JOB', value: "{{ instance.displayname }} - {{ environment }}"),
                 string(name: 'CHANNEL', value: "#dm-release"),
                 text(name: 'STAGE', value: "{{ environment }}"),
                 text(name: 'STATUS', value: status),
-                text(name: 'URL', value: "<${BUILD_URL}smoke_test_report|${BUILD_DISPLAY_NAME}>")
+                text(name: 'URL', value: "<${BUILD_URL}{{ instance.report_name|lower|replace(" ", "_") }}smoke_test_report|${BUILD_DISPLAY_NAME}>")
```
katstevens: Looking at the xml output for smoulder preview, this line resolves to 

```
text(name: 'URL', value: &quot;&lt;${BUILD_URL}smoulder_test_reportsmoke_test_report|${BUILD_DISPLAY_NAME}&gt;&quot;)
```

The quotes are fine but should `smoulder_test_reportsmoke_test_report` have an underscore in the middle?

job_definitions/smoke_tests.yml
```diff
@@ -1,42 +1,86 @@
...
               ]
       }
       node {
```
katstevens: The `stage('Run smoke tests')` stage in this section is still hardcoded - not sure if that's easy to change?

risicle:
These are almost all pushed now and appear to work. I just need to make another addition to the credentials repo before I can get it working for production.

PR #172, opened 2018-11-30
Enable SSH access logging on Jenkins box

lfdebrux:
This PR is part of this [ticket] to enable Jenkins SSH access logging.

Although the logging provided by the SSH daemon is less useful than logging at the load balancer in that it doesn't record client IP addresses, I still think it's useful to capture the SSH session details.

This PR adds an Ansible role to handle the SSHd configuration, and enables verbose logging to the file `/var/log/auth.log`.

Examples of log entries below:

```
Nov 22 11:15:18 ip-172-31-14-48 sshd[23708]: Postponed publickey for ubuntu from 172.31.6.132 port 4212 ssh2 [preauth]
Nov 22 11:15:18 ip-172-31-14-48 sshd[23708]: Accepted publickey for ubuntu from 172.31.6.132 port 4212 ssh2: RSA SHA256:deadbeef/redactedbeef
Nov 22 11:15:18 ip-172-31-14-48 sshd[23708]: pam_unix(sshd:session): session opened for user ubuntu by (uid=0)
Nov 22 11:15:18 ip-172-31-14-48 systemd-logind[1303]: New session 3797 of user ubuntu.
Nov 22 11:15:18 ip-172-31-14-48 systemd: pam_unix(systemd-user:session): session opened for user ubuntu by (uid=0)
Nov 22 11:15:18 ip-172-31-14-48 sshd[23708]: User child is on pid 23748
```

[ticket]: https://trello.com/c/S7iFoAU8/253-configure-jenkins-box-ssh-access-logging

### DIFF
--- a/playbooks/jenkins_playbook.yml
+++ b/playbooks/jenkins_playbook.yml
@@ -11,4 +11,5 @@
     sshd:
       ClientAliveCountMax: 80
       ClientAliveInterval: 45
+      LogLevel: VERBOSE
       PasswordAuthentication: no
--- a/playbooks/jenkins_playbook.yml
+++ b/playbooks/jenkins_playbook.yml
@@ -5,3 +5,10 @@
   become: yes
   roles:
     - jenkins
+    - role: willshersystems.sshd
+      tags: [sshd, jenkins]
+  vars:
+    sshd:
+      ClientAliveCountMax: 80
+      ClientAliveInterval: 45
+      PasswordAuthentication: no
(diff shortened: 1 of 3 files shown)


risicle:
Looks neat enough - isn't there at some point a generic `journald` entry extractor rather than per-daemon specific plugins?

lfdebrux:
> Looks neat enough - isn't there at some point a generic `journald` entry extractor rather than per-daemon specific plugins?

I don't think sshd is aligned that closely with journald or the systemd way of doing things.

My understanding is that sshd will not output these events unless the configuration option LogLevel is set to VERBOSE.

risicle:
*Oh*, well that's interesting.

PR #174, opened 2018-12-06
release pipeline: attempt to opportunistically launch smoulder tests on prod after successful pipeline release

risicle:
https://trello.com/c/y3rcfYeG/213-create-smoulder-tests-to-allow-more-thorough-tests-notably-antivirus-tests-to-be-run-on-production

The idea is that we get quicker feedback if something is wrong rather than
potentially having to wait up to an hour for them to be run automatically


risicle:
This seems to work - I've also removed the "check if already queued" mechanism because it turns out that jenkins does indeed have this behaviour by default.

PR #177, opened 2019-01-10
data_retention job: add $MAILCHIMP_API_TOKEN argument to enable mailchimp list cleaning

risicle:
This should enable https://trello.com/c/QROOUpoX, but I shouldn't apply it until https://github.com/alphagov/digitalmarketplace-scripts/pull/327 is merged.

### DIFF
--- a/job_definitions/data_retention.yml
+++ b/job_definitions/data_retention.yml
@@ -31,6 +31,10 @@
             FLAGS="--dry-run"
           fi
 
-          docker run -e DM_DATA_API_TOKEN_{{ environment|upper }} digitalmarketplace/scripts scripts/data-retention-remove-user-data.py '{{ environment }}' $FLAGS --verbose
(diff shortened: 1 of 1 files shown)


job_definitions/data_retention.yml
```diff
@@ -31,6 +31,6 @@
             FLAGS="--dry-run"
           fi
 
-          docker run -e DM_DATA_API_TOKEN_{{ environment|upper }} digitalmarketplace/scripts scripts/data-retention-remove-user-data.py '{{ environment }}' $FLAGS --verbose
+          docker run -e DM_DATA_API_TOKEN_{{ environment|upper }} digitalmarketplace/scripts scripts/data-retention-remove-user-data.py '{{ environment }}' --mailchimp-api-key=$MAILCHIMP_API_TOKEN $FLAGS --verbose
```
katstevens: Do we need to specify a `mailchimp_username` arg as well?

risicle:
Hmmmm having just merged & applied this I wonder if I should remove the mailchimp token from the staging & preview jobs. I guess their databases shouldn't be populated with real email addresses anyway but I wonder if I should put a warning anywhere...

PR #178, opened 2019-01-11
Change name of script called by notify-suppliers-of-dos-opportunities

lfdebrux:
We want to rename the script in the dmscripts repo, so we also need to update the job.

This PR depends on https://github.com/alphagov/digitalmarketplace-scripts/pull/328 and is part of https://trello.com/c/aRpNbH4o/309-rename-send-dos-opportunities-emailpy-to-notify-suppliers-something-somethingpy


lfdebrux:
Closing as we're not going to rename the script until it actually uses Notify

PR #179, opened 2019-01-14
Add job for building Docker base images periodically

lfdebrux:
We want to keep our Docker base images fresh so we're keeping up-to-date with the latest OS security fixes.

This PR adds a pipeline to build and push new versions of the Docker image every three days or so; it is very simple as it just uses the existing Makefile in the alphagov/digitalmarketplace-docker-base repo.


benvand:
Not sure if we need these bits when we're manually checking out repos:
https://github.com/alphagov/digitalmarketplace-jenkins/blob/master/job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml#L34
https://github.com/alphagov/digitalmarketplace-jenkins/blob/master/job_definitions/build_image.yml#L36

```
sh("git clean -fdx")
sh("git reset --hard ${RELEASE_NAME}")
```

benvand:
Looking in to it a bit more we could, by the looks of it, use

`git url: "git@github.com:alphagov/digitalmarketplace-${REPOSITORY}.git", branch: 'master', credentialsId: 'github_com_and_enterprise', wipe-workspace: true`

benvand:
https://docs.openstack.org/infra/jenkins-job-builder/project_workflow_multibranch.html#project_multibranch.git_scm

Hmm that's true by default

PR #187, opened 2019-01-28
Remove references to production users from envvars

lfdebrux:
Don't refer to users for tests as "production users". Basically a find-replace `s/PRODUCTION_//`.

Needed due to alphagov/digitalmarketplace-functional-tests#594 and alphagov/digitalmarketplace-visual-regression#20.


lfdebrux:
I've split it into two commits at @katstevens suggestion.

This way to get a seamless transition we 
1. push the first commit
2. merge the changes in the other PRs
3. wait five minutes or so
4. push the second commit.

PR #194, opened 2019-02-22
Fix generate counterpart signature pages

lfdebrux:
This job was causing issues because it was altering the global pyenv; I've fixed it so that it should now only be altering it's own venv.

In future this job should be changed to run in a Docker container, but it has a few peculiarities that mean that might be a bit trickier than normal.

### DIFF
--- a/job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
+++ b/job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
@@ -84,12 +84,13 @@
             currentBuild.result = 'FAILURE'
             echo "Error: ${err}"
             build job: 'notify-slack', parameters: [
-              string(name: 'USERNAME', value: 'jenkins'),
+              string(name: 'USERNAME', value: 'generate-upload-and-notify-counterpart-signature-pages'),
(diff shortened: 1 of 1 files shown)


job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
```diff
@@ -47,12 +47,11 @@
             stage('Generate') {
               dir('scripts') {
                 sh('pyenv local 3.6.2')
-                sh('[ -d venv ] || virtualenv venv')
-                sh('source ./venv/bin/activate')
-                sh('pip install --upgrade pip')
-                sh('pip install -r requirements.txt')
+                sh('[ -d venv ] || python3 -m venv venv')
+                sh('venv/bin/pip install --upgrade pip')
+                sh('venv/bin/pip install -r requirements.txt')
```
katstevens: `pip` or `pip3`?

lfdebrux: Better explicit than implicit 👍

lfdebrux: Fiexd

PR #195, opened 2019-03-04
377-aws-alarm-for-disk-space-on-jenkins

benvand:
Replace 'AWS logs' with 'amazon cloudwatch agent'
https://trello.com/c/PdXsLT6d/377-aws-alarm-for-disk-space-on-jenkins

### DIFF
--- a/playbooks/roles/jenkins/tasks/jenkins.yml
+++ b/playbooks/roles/jenkins/tasks/jenkins.yml
@@ -36,7 +36,9 @@
   vars:
     logrotate_scripts:
       - name: jenkins
-        path: /var/log/jenkins/access.log
+        paths:
+          - "/var/log/jenkins/access.log"
+          - "/var/log/amazon/amazon-cloudwatch-agent/amazon-cloudwatch-agent.log"
         options:
           - weekly
           - copytruncate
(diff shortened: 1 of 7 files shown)


playbooks/roles/jenkins/templates/amazon-cloudwatch-agent.json.j2
```diff
@@ -0,0 +1,53 @@
+{
+  "agent": {
+    "metrics_collection_interval": 60,
+    "region": "eu-west-1",
+    "logfile": "/opt/aws/amazon-cloudwatch-agent/logs/amazon-cloudwatch-agent.log",
```
lfdebrux: Does the metrics need to be logged to here? Could we put the log file in `/var/log` instead?

benvand: Good question.

It's actually the default location as detailed here:
screen shot 2019-03-04 at 15 12 27


Should I remove that line in the agent section, or move the logfile to `/var/log`?

playbooks/roles/jenkins/tasks/jenkins.yml
```diff
@@ -36,7 +36,9 @@
   vars:
     logrotate_scripts:
       - name: jenkins
-        path: /var/log/jenkins/access.log
+        paths:
+          - "/var/log/jenkins/access.log"
+          - "/var/log/amazon/amazon-cloudwatch-agent/amazon-cloudwatch-agent.log"
         options:
```
benvand: As per:
https://github.com/nickhammond/ansible-logrotate#role-variables

PR #210, opened 2019-04-01
Update jenkins to use mirror clone job_definitions/code_backups.yml

benvand:
* Rewrite of job to properly mirror repos

https://trello.com/c/0O4zrb6x/442-code-backups

### DIFF
--- a/job_definitions/code_backups.yml
+++ b/job_definitions/code_backups.yml
@@ -33,27 +33,25 @@
 {% for package in packages %}
 - job:
     name: 'backup-{{ package }}'
-    display-name: "Backup {{ package }} Github repo to AWS CodeCommit"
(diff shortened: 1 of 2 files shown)


job_definitions/code_backups.yml
```diff
@@ -33,27 +33,25 @@
 {% for package in packages %}
 - job:
     name: 'backup-{{ package }}'
-    display-name: "Backup {{ package }} Github repo to AWS CodeCommit"
+    display-name: "[code-backup] {{ package }}"
```
katstevens: Could we use something other than `[`? This pushes the pipeline jobs down from the top of the main Jenkins list 😿

benvand: Done 👍

PR #212, opened 2019-04-15
Reenable exporting data

lfdebrux:
We've given the gdrive api a bit of a rest, and we've also stopped uploading large files, so let's reenable the data export and see whether the limit is still breached constantly.

### DIFF
--- a/job_definitions/export_data.yml
+++ b/job_definitions/export_data.yml
@@ -1,12 +1,10 @@
 {% set environments = ['preview', 'production'] %}
-{% set disabled = 'true' %}
 ---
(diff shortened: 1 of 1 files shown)


job_definitions/export_data.yml
```diff
@@ -1,5 +1,4 @@
 {% set environments = ['preview', 'production'] %}
-{% set disabled = 'true' %}
```
katstevens: We might have to explicitly set `disabled = 'false'` or remove the usage of it below.

lfdebrux: Good point, fixed now

PR #216, opened 2019-05-10
build docker image job: add options HEAD_BRANCH and MARK_LATEST

risicle:
In the lead up to investigating https://trello.com/c/skD666lL

These should allow the job to build experimental off-branch images without interfering with the regular release pipeline. Right?

What do people think? I don't know what I'm doing.

### DIFF
--- a/job_definitions/build_image.yml
+++ b/job_definitions/build_image.yml
@@ -42,6 +42,12 @@
               }
 
               stage('Prepare') {
+                  if (RELEASE_NAME ==~ /release-\d+/ && HEAD_BRANCH != "master") {
+                    input("WARNING: It is probably a bad idea to create a release- tagged image from a non-master HEAD_BRANCH!")
+                  }
+                  if (MARK_LATEST.toString() == "true" && HEAD_BRANCH != "master") {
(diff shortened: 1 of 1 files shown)


risicle:
(hold on I haven't actually injected the variable into the script...)

risicle:
FWIW I're run this once and it does appear to work

katstevens:
I think I agree with Ben that there should be more of a "⚠️ Are you sure? ⚠️" around this if the branch != `master`, but otherwise I think it's OK.

risicle:
Yeah ok I'll see how far I can push my non-existent dsl skillz

lfdebrux:
I have an alternate PR that I have been sitting on for a while if anyone is interested

risicle:
> I have an alternate PR that I have been sitting on for a while if anyone is interested

Interested!

risicle:
@katstevens the only way I can see to add an "are you sure?" to a job is to convert it to a pipeline. And I'm not sure that's something we really want to do too much to jobs that don't need it otherwise.

katstevens:
🤔 I thought this job was already a pipeline?

risicle:
Oh... yes ... it is.

Let... me... see...

risicle:
Ok whaddya all think? I've added dire warnings.

PR #224, opened 2019-06-07
Use all YAML files for Jenkins in credentials repo

lfdebrux:
Allows credentials files added by https://github.com/alphagov/digitalmarketplace-credentials/pull/221 to be picked up by Ansible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,9 @@ jenkins: requirements ## Run Jenkins playbook
 	JENKINS_VARS_FILE=$$(mktemp) ;\
 	PRIVATE_KEY_FILE=$$(mktemp) ;\
 	trap 'rm $$JENKINS_VARS_FILE $$PRIVATE_KEY_FILE' EXIT ;\
-	${DM_CREDENTIALS_REPO}/sops-wrapper -d ${DM_CREDENTIALS_REPO}/jenkins-vars/jenkins.yaml > $$JENKINS_VARS_FILE ;\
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -47,7 +47,9 @@ jenkins: requirements ## Run Jenkins playbook
 	JENKINS_VARS_FILE=$$(mktemp) ;\
 	PRIVATE_KEY_FILE=$$(mktemp) ;\
 	trap 'rm $$JENKINS_VARS_FILE $$PRIVATE_KEY_FILE' EXIT ;\
-	${DM_CREDENTIALS_REPO}/sops-wrapper -d ${DM_CREDENTIALS_REPO}/jenkins-vars/jenkins.yaml > $$JENKINS_VARS_FILE ;\
+	for varfile in ${DM_CREDENTIALS_REPO}/jenkins-vars/*.yaml; do \
```
katstevens: Should we allow `.yml` as well as `yaml` for flexibility, or should we enforce consistency? (I am on the fence...)

lfdebrux: Google says... https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml?rq=1

lfdebrux: I prefer consistency

lfdebrux: Also the only yaml files in the relevant folder so far are .yaml so hopefully that means that we shouldn't need the flexibility

katstevens: Sounds good to me!

PR #227, opened 2019-06-07
Add job to validate CloudTrail logs periodically

lfdebrux:
Implements https://trello.com/c/Ijzuk0qY/508-log-file-validation-job, needs https://github.com/alphagov/digitalmarketplace-credentials/pull/221 and https://github.com/alphagov/digitalmarketplace-scripts/pull/387 to be merged to work.

This PR creates a job that will hourly test the last three days of CloudTrail logs for each of our four AWS accounts. This schedule was suggested by @agalmatis but isn't fixed in stone.

The job relies on being able to assume the `cloudtrail-validate-logs` role created in https://github.com/alphagov/digitalmarketplace-aws/pull/572. The way that this has been achieved is by using the fact that Jenkins runs on AWS EC2; this PR adds a config file for the AWS cli with the role ARNs and specifies that AWS should use its EC2 instance profile to assume these roles. This means we don't need to worry about MFA when running the validate-cloudtrail-logs script on Jenkins with the correct profile. A new Ansible task takes care of adding this config file.

### DIFF
--- a/job_definitions/validate_cloudtrail_logs.yml
+++ b/job_definitions/validate_cloudtrail_logs.yml
@@ -1,0 +1,92 @@
+---
+- job:
+    name: validate-cloudtrail-logs
+    project-type: freestyle
+    display-name: Validate logs stored by Amazon CloudWatch in S3.
+    parameters:
+      - choice:
+          name: ACCOUNT
+          choices:
+            - main
+            - development
+            - production
+            - backups
+      - string:
+          name: START_FROM
+          default: "3 days ago"
+          description: Logs delivered on or after the specified time will be validated. Can be in any format accepted by GNU date.
+      - string:
+          name: UP_TO
+          default: "now"
+          description: Logs delivered on or before the specified time will be validated. Can be in any format accepted by GNU date.
+      - bool:
+          name: VERBOSE
+          default: false
+          description: Print verbose log validation information.
+    builders:
+      - shell: |
+          docker run --rm \
+            -v /var/lib/jenkins/digitalmarketplace-credentials:/digitalmarketplace-credentials \
+            -e DM_CREDENTIALS_REPO=/digitalmarketplace-credentials \
+            -e VERBOSE=$VERBOSE \
+            digitalmarketplace/scripts \
+            scripts/validate-cloudtrail-logs.sh $ACCOUNT $START_FROM $UP_TO
+
+- job:
+    name: validate-cloudtrail-logs-periodic
+    project-type: pipeline
+    display-name: Validate the last 3 days of logs every hour or so.
+    triggers:
+      - timed: '@hourly'
+    dsl: |
+
+      def notify_slack(accounts, status) {
+        build job: "notify-slack",
+        parameters: [
+          string(name: 'USERNAME', value: "validate-cloudtrail-logs"),
+          string(name: 'ICON', value: ":red_circle:"),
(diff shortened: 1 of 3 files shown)


playbooks/roles/jenkins/templates/aws_config.j2
```diff
@@ -0,0 +1,7 @@
+{% for role in assume_roles %}
```
benvand: Not sure you're going to be able to assume this role without adding jenkins here:
https://github.com/alphagov/digitalmarketplace-aws/blob/master/terraform/modules/cloudtrail/modules/cloudtrail-validate-logs-role/main.tf#L4-L3

benvand: I don't massively dislike the paradigm but we could just add the policy statements here
https://github.com/alphagov/digitalmarketplace-aws/blob/master/terraform/modules/cloudtrail/modules/cloudtrail-validate-logs-role/main.tf#L19
 to the jenkins instance profile here:
https://github.com/alphagov/digitalmarketplace-aws/blob/master/terraform/accounts/main/jenkins_instance_profile.tf#L39

benvand: _I think_

lfdebrux: > I don't massively dislike the paradigm but we could just add the policy statements here
> https://github.com/alphagov/digitalmarketplace-aws/blob/master/terraform/modules/cloudtrail/modules/cloudtrail-validate-logs-role/main.tf#L19
> to the jenkins instance profile here:
> https://github.com/alphagov/digitalmarketplace-aws/blob/master/terraform/accounts/main/jenkins_instance_profile.tf#L39

The reasoning was that it makes it easier for anyone to run the task, not just the Jenkins box

job_definitions/validate_cloudtrail_logs.yml
```diff
@@ -0,0 +1,92 @@
...
+          description: Print verbose log validation information.
+    builders:
+      - shell: |
+          docker run --rm \
+            -v /var/lib/jenkins/digitalmarketplace-credentials:/digitalmarketplace-credentials \
+            -e DM_CREDENTIALS_REPO=/digitalmarketplace-credentials \
+            -e VERBOSE=$VERBOSE \
+            digitalmarketplace/scripts \
+            scripts/validate-cloudtrail-logs.sh $ACCOUNT $START_FROM $UP_TO
+
```
benvand: Think it would be cool to do this as a loop for the different accounts. On the off chance it fails then it won't fail for all and will indicate the correct account.

playbooks/roles/jenkins/templates/aws_config.j2
lfdebrux: > Not sure you're going to be able to assume this role without adding jenkins here:
> https://github.com/alphagov/digitalmarketplace-aws/blob/master/terraform/modules/cloudtrail/modules/cloudtrail-validate-logs-role/main.tf#L4-L3

Oh yeah haven't got around to that yet 😆

job_definitions/validate_cloudtrail_logs.yml
agalamatis: @benvand There are two kind of "failures", one is when validation is not performed for any reason (permissions, connectivity, etc.), while the other is when validation returns at least one invalid log or digest file. In both cases these are caught and stored below, telling us which account failed for us to check manually. To my understanding if one stage fails the rest will be checked, but I could be wrong.

PR #230, opened 2019-06-13
One Punch Jenkins

lfdebrux:
Refactor Ansible so that it will build Jenkins with one round of Ansible.

This has a load of updates/fixes that I found necessary when I had to rebuild Jenkins recently. I also did some general tidying.

I have tested it by running it against a blank EC2 instance.


playbooks/roles/jenkins/tasks/05_tools.yml
```diff
@@ -112,13 +119,22 @@
...
-    owner: jenkins
-    group: jenkins
-    state: directory
-    recurse: yes
+# TODO: Stop installing PhantomJS like this when there is a better
+#       apt source available.
+- name: Download PhantomJS
+  get_url:
+    url: "https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linux-x86_64.tar.bz2"
+    dest: /usr/local/src/phantomjs-2.1.1-linux-x86_64.tar.bz2
```
benvand: It's worth checking the integrity of this downloaded binary with the md5 checksum here:
http://phantomjs.org/download.html
We've done this somewhere else but I can't find the example

benvand: This _might_ help

https://github.com/alphagov/digitalmarketplace-jenkins/blob/master/playbooks/roles/jenkins/tasks/sops.yml#L7

lfdebrux: Done

PR #246, opened 2019-07-09
Use Dockerfile for building manual

lfdebrux:
https://trello.com/c/I6n0Dx9k/930-publish-manual-job-needs-love

Requires https://github.com/alphagov/digitalmarketplace-manual/pull/197 to be merged and the docker repo `digitalmarketplace-manual` to be created on Docker Hub.

This PR changes the manual job to build the html in a Docker container with Sphinx already installed. This means we don't need a virtualenv or pyenv for this job.

### DIFF
--- a/job_definitions/manual.yml
+++ b/job_definitions/manual.yml
@@ -15,8 +15,13 @@
           cron: "H/2 * * * *"
     builders:
       - shell: |
-          [ -x /tmp/dm-manual-venv/bin/pip ] || virtualenv /tmp/dm-manual-venv
-          . /tmp/dm-manual-venv/bin/activate
+          docker pull digitalmarketplace/manual || true
+          docker build --pull -t digitalmarketplace/manual .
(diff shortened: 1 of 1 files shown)


job_definitions/manual.yml
```diff
@@ -15,8 +15,13 @@
...
+          docker push digitalmarketplace/manual
 
-          pip install -r requirements.txt
-          make clean html pages
+          docker run -it --rm \
+            -v $(pwd):/app \
+            digitalmarketplace/manual \
+            make clean html
+          
+          make pages
```
benvand: `make pages` runs `make html` which runs `${VIRTUALENV_ROOT}/bin/sphinx-build -b html ARGS`

So it'll still need a venv here I think

lfdebrux: Good spot, I've added a commit to https://github.com/alphagov/digitalmarketplace-manual/pull/197 to remove the `make html` line from `buildpages`

PR #250, opened 2019-07-11
Don't use disabled in job defs

benvand:
Remove explicitly setting disabled in the jobs, this overrides the disabled flag set in job_definitions/_defaults.yml which is bad

This flag should not be overrider as jobs are disabled when jenkins is built to prevent unexpected runs

### DIFF
--- a/job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
+++ b/job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
@@ -1,78 +1,93 @@
-{% set environments = ['production'] %}
-{% set frameworks = [('DOS3', 'digital-outcomes-and-specialists-3'), ('G11', 'g-cloud-11')]%}
+{# frameworks = (acronym, framework_slug, enabled?) #}
+{% set frameworks = [
+  ('DOS3', 'digital-outcomes-and-specialists-3', True),
+  ('G10', 'g-cloud-11', False),
+  ('G11', 'g-cloud-11', True),
+]%}
 ---
-{% for environment in environments %}
-{% for acronym, framework in frameworks %}
 - job:
-    name: 'generate-upload-and-notify-counterpart-signature-pages-{{ environment }}-{{ acronym }}'
-    display-name: 'Generate, Upload and Notify Counterpart Signature Pages - {{ environment }}-{{ acronym }}'
-    project-type: pipeline
+    name: generate-upload-and-notify-counterpart-signature-pages
+    display-name: Generate, upload and notify counterpart signature pages
     description: |
-        This job will run a series of scripts to
+        <p>This job will run a series of scripts to</p>
         <ol>
           <li>Generate framework agreement counterpart signature pages for suppliers approved for countersignature but
               without a countersigned agreement</li>
           <li>Upload the generated files to S3 and set the path in the relevant FrameworkAgreement record</li>
           <li>Send an email to all active users on the supplier account that the counterpart file has been uploaded</li>
         </ol>
+    project-type: pipeline
     concurrent: false
-    triggers:
-      - timed: 'H 1 * * *'
     properties:
       - build-discarder:
           days-to-keep: 7
           artifact-days-to-keep: 7
+    parameters:
+      - choice:
+          name: ENVIRONMENT
+          choices:
+            - preview
+            - staging
+            - production
+      - string:
+          name: FRAMEWORK
+      - bool:
+          name: DRY_RUN
+          default: false
     dsl: |
       node {
           try {
-            stage('Prepare') {
-              dir('scripts/generated-countersignature-pdfs') {
-                deleteDir()
+            stage('Create data directory') {
+              sh('rm -rf ./data && mkdir data')
+            }
+            stage('Clone agreements') {
+              dir('digitalmarketplace-agreements') {
+                git url: 'git@github.com:alphagov/digitalmarketplace-agreements.git', branch: 'master', credentialsId: 'github_com_and_enterprise', poll: true
               }
-              dir('scripts') {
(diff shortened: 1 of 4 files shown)


lfdebrux:
Hmm, that is a good point... However it is annoying 😆 How useful is it to be able to toggle disabled like this? Thoughts @katstevens?

katstevens:
Hmm. It's pretty useful for when we're trying to setup new framework jobs in advance. For example the performance platform jobs will fail if the framework isn't open for applications (this is probably a good thing - we should know about it if the framework status isn't `open` for whatever reason). We could potentially change the scripts themselves to be more robust but that'll need some work.

lfdebrux:
Okay I've added some commits that I think make a reasonable workaround

job_definitions/stats_snapshots.yml
```diff
@@ -1,14 +1,17 @@
...
-    name: 'snapshot-stats-{{ framework }}-{{ environment }}'
-    disabled: {{ disabled }}
-    display-name: "Snapshot {{ framework }} stats - {{ environment }}"
-    project-type: freestyle
+    name: stats-snapshot-{{ framework }}-{{ environment }}
+    display-name: Snapshot stats for {{ framework }} - {{ environment }}
     description: Creates an hourly snapshot of framework stats and stores it in the API audit event
+    {% if not enabled %}
+    disabled: true
+    {% endif %}
```
katstevens: I think this should be fine. Though be careful with whitespace/indentation of `{% for %}` loops in Jenkins jobs templates, we've come unstuck with those before 😿

lfdebrux: Eep good point

lfdebrux: I'll change it

benvand:
@katstevens I can't approve this either :/

job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
```diff
@@ -21,60 +16,60 @@
...
+            - staging
+            - production
+      - string:
+          name: FRAMEWORK
     dsl: |
       node {
           try {
-            stage('Prepare') {
-              dir('scripts/generated-countersignature-pdfs') {
-                deleteDir()
```
katstevens: Are we no longer explicitly cleaning up this folder? We have a new data folder cleaned & created below - is that just for cloning repos? Where do the PDFs go?

katstevens:
I am happy with the changes to the stats jobs, but the generate-and-upload job is more of a major change (and is an essential job for G11 suppliers right now). Is there a way we can test it?

job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
lfdebrux: PDFs go to `data/generated-countersignature-pdfs`, so it gets cleaned up by the `rm -rf` below

lfdebrux:
> I am happy with the changes to the stats jobs, but the generate-and-upload job is more of a major change (and is an essential job for G11 suppliers right now). Is there a way we can test it?

I've added a dry-run flag and am going to test now by uploading a preview version of the job to Jenkins.

lfdebrux:
Have tested the counterpart signatures job on Jenkins, works now

job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
```diff
@@ -96,6 +97,34 @@
...
+    display-name: 'Generate, upload and notify counterpart signature pages for {{', '.join(s) for s,_,enabled if enabled}}'
+    project-type: pipeline
+    description: |
+        This job will run a series of scripts to
+        <ol>
+          <li>Generate framework agreement counterpart signature pages for suppliers approved for countersignature but
+              without a countersigned agreement</li>
+          <li>Upload the generated files to S3 and set the path in the relevant FrameworkAgreement record</li>
+          <li>Send an email to all active users on the supplier account that the counterpart file has been uploaded</li>
+        </ol>
```
katstevens: We have the same description on this `-periodic` job and the one above - we should probably mention that this one calls the other?

job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
```diff
@@ -96,6 +97,34 @@
             ]
           }
       }
+  
+- job:
+    name: 'generate-upload-and-notify-counterpart-signature-pages-periodic'
+    display-name: 'Generate, upload and notify counterpart signature pages for {{', '.join(s) for s,_,enabled if enabled}}'
```
katstevens: Can we include the `<stage>` in the `display-name` if it's not a parameter we can set when building the job?

job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
katstevens: Aren't we deleting `scripts/generated-countersignature-pdfs` in the `Cleanup` stage?

lfdebrux: Ah good spot!!

lfdebrux:
Okay I think I've ironed out all the bugs?

katstevens:
Just thought of another thing (sorry! 😹) Will these new jobs need to be included in the config list to show up on the Framework Lifecycle jobs tab?

lfdebrux:
> Just thought of another thing (sorry! ) Will these new jobs need to be included in the config list to show up on the Framework Lifecycle jobs tab?

Done

lfdebrux:
> I am finally happy 😸

Thanks for being so diligent!

PR #254, opened 2019-07-12
Pin cf-cli to 0.63.*

lfdebrux:
cf-cli analyses manifest files using multiple passes of the file
descriptor, which means that using process substitution (`<()`) with
generate manifest doesn't always work (c.f. #74).

The long term fix is to rework our deploy scripts so we don't need to
use process substitution with the manifests, but for now the workaround
is to downgrade cf-cli.


risicle:
https://nvd.nist.gov/vuln/detail/CVE-2019-3781

lfdebrux:
Yeah it's not ideal

PR #270, opened 2019-09-25
Update DOS opportunities email job in Jenkins tabs

katstevens:
https://trello.com/c/58q6lRKF/33-update-dos4-regular-email-jobs

The DOS opportunities email job got renamed in https://github.com/alphagov/digitalmarketplace-jenkins/pull/268. The config needs to be updated as well, to show the new job name in the correct tab and on the build monitor view.


katstevens:
Exactly! We should no longer have to update this job every year 🎉

PR #274, opened 2019-10-02
generate_upload_and_notify_counterpart_signature_pages: add `wheel` package to venv

risicle:
This allows pip to cache compiled packages and avoid rebuilding them all on every invocation. Which seems like it occupies a significant amount of the runtime of this script.

Also stop this script annoying `#dm-2ndline` for non-production failures.

Need I say, this is actually already deployed and working.

### DIFF
--- a/job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
+++ b/job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
@@ -56,6 +56,7 @@
 
                   make virtualenv
                   ./venv/bin/pip install --upgrade pip
+                  ./venv/bin/pip install wheel  # this allows pip to cache compiled packages instead of rebuilding them all on every invocation
                   make requirements
                 ''')
               }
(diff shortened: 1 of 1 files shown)


job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
```diff
@@ -102,7 +102,7 @@
               string(name: 'ICON', value: ':alarm_clock:'),
               string(name: 'STAGE', value: "${ENVIRONMENT}"),
               string(name: 'STATUS', value: 'FAILED'),
-              string(name: 'CHANNEL', value: '#dm-2ndline'),
+              string(name: 'CHANNEL', value: ENVIRONMENT == 'production' ? '#dm-2ndline' : '#dm-release'),
```
katstevens: Just to check, are we using ternary here rather than the usual jinja logic `{{'#dm-2ndline' if environment == 'production' else '#dm-release'}}` as we don't have `environment` available to us?

risicle: Err no, I just forgot we had jinja available to us... it would probably be better to do it that way if poss.

PR #276, opened 2019-10-21
Add visual regression tests job to staging

lfdebrux:
Ticket: https://trello.com/c/LxcD30hu/55-add-visual-regression-tests-against-staging-to-release-pipeline

We want to be able to check whether changes we are releasing will cause visual regressions.

This commit adds a visual-regressions-staging job, and sets it to be automatically run as part of the release pipeline for an app.

We also make sure visual regressions are run against staging when restoring a database dump. ~I copied the function from the release pipeline that asks for user input if the visual regression tests fail. Since we're expecting the visual regression tests to fail this does mean we'll have to check them weekly to approve them; alternatively we could approve automatically.~ I've added a function that runs the visual regression tests and automatically approves the changes.

### DIFF
--- a/job_definitions/clean_and_apply_db_dump.yml
+++ b/job_definitions/clean_and_apply_db_dump.yml
@@ -13,6 +13,43 @@
 {% endif %}
     dsl: |
 
+      def run_visual_regression_tests(stage) {
+        def job_name = "visual-regression-${stage}"
+        def vr_job = null
+
+        node {
+          vr_job = build job: job_name, parameters: [
+            string(name: "COMMAND", value: "test"),
+          ], propagate: false
+        }
+
+        def vr_job_result = vr_job.result
+        def vr_job_number = vr_job.number
+        vr_job = null
+
+        if (vr_job_result == 'SUCCESS') {
+          return true
+        } else {
+          echo "Visual regression tests build result: ${vr_job_result}"
+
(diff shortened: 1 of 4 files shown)


katstevens:
I'm leaning towards auto-approval after the db-clean job. It will almost certainly 'fail' every week, but unless we had to abandon something mid-deploy on Friday afternoon, there shouldn't be any non-data changes on staging. The likelihood of us forgetting to approve the pipeline on a Monday morning (or just mindlessly approving without checking all the failures thoroughly) seems high. If we keep the manual approval step just for release failures, then a) someone should be paying close attention anyway b) there should be less noise so we can see the intended changes and pick up on any unexpected changes.

lfdebrux:
Okay I've updated the changes in line with your comments @katstevens

job_definitions/clean_and_apply_db_dump.yml
```diff
@@ -13,6 +13,43 @@
...
+        def vr_job_result = vr_job.result
+        def vr_job_number = vr_job.number
+        vr_job = null
+
+        if (vr_job_result == 'SUCCESS') {
+          return true
+        } else {
+          echo "Visual regression tests build result: ${vr_job_result}"
+
+          echo "Assuming that these changes are due to databse restore. Approving."
```
katstevens: Small typo `databse`

PR #283, opened 2019-11-21
add rotate-ft-account-passwords job

risicle:
https://trello.com/c/AOxxcCSy

This is a fairly basic initial stab at this pipeline which we should probably discuss/flesh out.

The reason this is a bit of a tricky process is the time between merging the PR/restarting jenkins/syncing the passwords is a bit of a "critical section". We can't use a graceful switchover technique similar to the other token rotation scripts because DMP accounts don't support multiple passwords.

Really the best thing to do is disable the functional/smoke/smoulder tests during these phases then re-enable them. But should we hint at this strongly in the pipeline? Put _another_ "stop and do this" stage in?

It does actually look like it would be possible to disable jobs _from_ another job, so hypothetically we could do it all automatically... but do we dare get into that kind of stuff?

### DIFF
--- a/job_definitions/rotate_ft_account_passwords.yml
+++ b/job_definitions/rotate_ft_account_passwords.yml
@@ -1,0 +1,93 @@
+---
+- job:
+    name: "rotate-ft-account-passwords"
+    display-name: "Rotate functional test account passwords"
+    project-type: pipeline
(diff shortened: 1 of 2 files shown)


risicle:
I've now done a full run of this pipeline for preview, it worked.

risicle:
Someone with better bash should have _written_ the other PR

PR #284, opened 2019-12-12
rotate-ft-account-passwords job: fix by adding update-credentials step and changing suggested ansible command

risicle:
https://trello.com/c/VKQKtzQW

Two things were wrong with the previous incarnation:

 - It prompted the user to do a `make jenkins TAGS=config` to update Jenkins' global environment variables. The problem is, the FT credentials aren't held in the global environment - they're in the job definitions themselves. So instead the user should be prompted to use `TAGS=jobs`.
 - It failed to update Jenkins' credentials repo before attempting to set the new passwords in the database, so ended up setting the existing ones.

The combination of these two problems however meant that everything continued to work and everything appeared to be fine... except the values now held in the credentials repo didn't reflect reality.

### DIFF
--- a/job_definitions/rotate_ft_account_passwords.yml
+++ b/job_definitions/rotate_ft_account_passwords.yml
@@ -51,12 +51,15 @@
           }
           milestone()
         }
-        stage("Update Jenkins Config") {
+        stage("Update Jenkins Credentials") {
+          build job: "update-credentials"
+        }
+        stage("Update Jenkins Jobs") {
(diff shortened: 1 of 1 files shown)


risicle:
I've pushed another change to this - commenting out the `staging` and `preview` options for `STAGE` because, clearly, changing these causes trouble. Why not just remove the option? I want to make it clear that they do still actually _work_ if anyone wants to uncomment them.

PR #285, opened 2019-12-16
clean and apply db dump job: add functional test run before visual regression acceptance

risicle:
The functional tests make a few small changes to the state of the database which otherwise cause the newly-accepted visual regression result to fail as soon as the FTs are run after import.

I haven't tested this. Maybe I should.

### DIFF
--- a/job_definitions/clean_and_apply_db_dump.yml
+++ b/job_definitions/clean_and_apply_db_dump.yml
@@ -180,6 +180,11 @@
 
 
 {% if environment in ['preview', 'staging'] %}
+              stage('Run functional tests') {
+                # running the functional tests should get the database in the right state for VR tests to produce identical results
+                build job: "functional-tests-{{ environment }}"
+              }
+
               stage('Run visual regression tests') {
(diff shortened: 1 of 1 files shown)


lfdebrux:
Have you tested this yet?

risicle:
So long ago I can't actually remember...

lfdebrux:
It might be nice to catch any errors from the functional tests, so it's clear that if they do fail that it's not the DB restore that has failed

risicle:
Does `stage(...)` not do that?

Or do you mean catch & discard FT errors?

lfdebrux:
> Does `stage(...)` not do that?
> 
> Or do you mean catch & discard FT errors?

I was more thinking about Slack notifications. I think catching and discarding is probably fine? Not sure it matters that much though.

risicle:
Oh yes. Hmm - might do that...

risicle:
Uhh.. but I don't have time to dig in and figure out pipeline dsl exception handling and I'd like this to be sorted out for the weekend, so...

PR #286, opened 2020-01-07
Jenkins job for framework clarification question notification script

katstevens:
https://trello.com/c/xbqxyfNV/210-notify-suppliers-about-new-framework-clarification-questions-via-the-admin

Job for running the new framework clarification question email script: https://github.com/alphagov/digitalmarketplace-scripts/pull/459.

Assumes that the job will be manually triggered for now. We can update the job later if we introduce any sort of S3/etag hook.

I've hardcoded some settings for the time being:
- Set `disabled` to True for now as well (we can enable it once the framework opens)
- Included `preview` environment for testing purposes (to be removed later, so I've only included the production job in the listings)
- Set `framework_slug` to `g-cloud-12`. We're only going to set this once per framework - having to type it in as a parameter on each run would be tedious, and it's very unlikely that our service will have multiple frameworks open at once in the near future.

### DIFF
--- a/job_definitions/notify_suppliers_of_framework_clarification_questions_answers.yml
+++ b/job_definitions/notify_suppliers_of_framework_clarification_questions_answers.yml
@@ -1,0 +1,52 @@
+{% set environments = ['preview', 'production'] %}
+{% set notify_template_id = 'ed9944eb-112e-4f04-a4bb-19f397550ffe' %}
+{% set framework_slug = 'g-cloud-12' %}
+---
+{% for environment in environments %}
+- job:
+    name: "notify-suppliers-of-framework-clarification-questions-answers-{{ environment }}"
+    display-name: "Notify suppliers of new framework clarification questions and answers - {{ environment }}"
+    project-type: freestyle
+    disabled: true
+    description: |
+      Send email notifications to supplier users about new framework clarification questions and answers.
+      This job is triggered manually, after a framework manager admin has uploaded a new set of clarification PDFs.
+    parameters:
+      - string:
+          name: RESUME_RUN_ID
+          description: "UUID of a previously failed run to use for resending (this will skip any users who were successfully emailed)"
+      - bool:
+          name: DRY_RUN
+          default: false
+          description: "List notifications that would be sent without sending the emails"
+    scm:
+      - git:
+          url: git@github.com:alphagov/digitalmarketplace-scripts.git
+          credentials-id: github_com_and_enterprise
+          branches:
+            - master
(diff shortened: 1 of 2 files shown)


katstevens:
A dry-run on preview went blue 🎉

job_definitions/notify_suppliers_of_framework_clarification_questions_answers.yml
```diff
@@ -0,0 +1,52 @@
+{% set environments = ['preview', 'production'] %}
+{% set notify_template_id = 'ed9944eb-112e-4f04-a4bb-19f397550ffe' %}
+{% set framework_slug = 'g-cloud-12' %}
```
lfdebrux: Could these be job parameters instead of being hardcoded into the template?

job_definitions/notify_suppliers_of_framework_clarification_questions_answers.yml
```diff
@@ -0,0 +1,52 @@
...
+          - project: notify-slack
+            condition: UNSTABLE_OR_WORSE
+            predefined-parameters: |
+              USERNAME=framework-question-and-answers
+              JOB=Notify suppliers of new framework clarification questions and answers {{ environment }}
+              ICON=:frame_with_picture:
+              STAGE={{ environment }}
+              STATUS=FAILED
+              URL=<${BUILD_URL}consoleFull|#${BUILD_NUMBER}>
+              CHANNEL=#dm-2ndline
```
lfdebrux: Can we include the framework name for information in the Slack message somehow?

job_definitions/notify_suppliers_of_framework_clarification_questions_answers.yml
katstevens: Yep, defaults could work. I'll have a think...

job_definitions/notify_suppliers_of_framework_clarification_questions_answers.yml
katstevens: Good idea!

katstevens: Done

job_definitions/notify_suppliers_of_framework_clarification_questions_answers.yml
katstevens: Have pushed a commit 👍

katstevens:
Thanks both - repushing the Jenkins job now, which will create the prod version as well.

PR #293, opened 2020-02-17
Add random password for PostgreSQL container

lfdebrux:
Ticket: https://trello.com/c/VBYabYVZ/1436-db-restore-on-staging-failed

Our clean_and_apply_db_dump job on Jenkins has stopped working because the postgres Docker image expects an environment variable POSTGRES_PASSWORD.

This commit adds a command to the prepare step to generate a random password and adds it to the environment. We also set PGPASSWORD in the environment for the psql command.

Depends on alphagov/digitalmarketplace-aws#628


lfdebrux:
Tested on Jenkins

PR #303, opened 2020-03-17
New jenkins job for generic framework emails

katstevens:
https://trello.com/c/RBJb5rfV/356-jenkinsify-remaining-framework-emails-2

A new job to allow developers to trigger framework emails for a number of (pre-defined) templates. This includes the 'new framework clarification questions have been published' email, which already has its own Jenkins job (which we can delete after this job has been tested/merged).

This group of emails will always be sent to the same group of users - suppliers who have started an application for the framework in question. Other framework-related emails get sent to different user groups so are not included here.

See https://github.com/alphagov/digitalmarketplace-scripts/pull/475 for the new script, based on the existing clarification questions job.

### DIFF
--- a/job_definitions/notify_suppliers_of_framework_application_events.yml
+++ b/job_definitions/notify_suppliers_of_framework_application_events.yml
@@ -1,0 +1,79 @@
+{% set environments = ['preview', 'production'] %}
+{% set notify_template_map = {
+    '21_new_clarification_questions_answers': 'ed9944eb-112e-4f04-a4bb-19f397550ffe',
+    '22_clarification_questions_close_next_week': '119dde9a-1e44-4b1d-9224-2f7020a74f9f',
+    '23_clarification_questions_close_today': '1464000b-1027-4fa5-97fe-7280ab9daf1d',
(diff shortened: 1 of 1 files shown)


job_definitions/notify_suppliers_of_framework_application_events.yml
```diff
@@ -0,0 +1,79 @@
...
+      - shell: |
+          if [ -n "$RESUME_RUN_ID" ]; then
+            FLAGS="--resume-run-id=$RESUME_RUN_ID"
+          fi
+
+          if [ "$DRY_RUN" = "true" ]; then
+            FLAGS="$FLAGS --dry-run"
+          fi
+
+          docker run -e DM_DATA_API_TOKEN_{{ environment|upper }} digitalmarketplace/scripts scripts/notify-suppliers-of-framework-application-event.py {{ environment }} $FRAMEWORK_SLUG $NOTIFY_API_TOKEN {{ notify_template_map[EMAIL_TYPE] }} $FLAGS
```
lfdebrux: Might want to break this up into several lines, also it's probably safer to wrap the different parameters in quote marks

PR #306, opened 2020-03-20
Add release notification to notify-suppliers-of-framework-application-event

lfdebrux:
This job will usually be triggered manually, so a success notification would be nice.

time

### DIFF
--- a/job_definitions/notify_suppliers_of_framework_application_events.yml
+++ b/job_definitions/notify_suppliers_of_framework_application_events.yml
@@ -51,6 +51,7 @@
               USERNAME=framework-application-event-email
               JOB=Notify suppliers of new ${FRAMEWORK_SLUG} framework event - {{ environment }}
               ICON=:frame_with_picture:
+              PROJECT=${NOTIFY_TEMPLATE_NAME}
               STAGE={{ environment }}
               STATUS=FAILED
               URL=<${BUILD_URL}consoleFull|#${BUILD_NUMBER}>
@@ -61,6 +62,7 @@
               USERNAME=framework-application-event-email
               JOB=Notify suppliers of new ${FRAMEWORK_SLUG} framework event - {{ environment }}
               ICON=:time:
+              PROJECT=${NOTIFY_TEMPLATE_NAME}
               STAGE={{ environment }}
               STATUS=SUCCESS
               URL=<${BUILD_URL}consoleFull|#${BUILD_NUMBER}>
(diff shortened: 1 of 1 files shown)


job_definitions/notify_suppliers_of_framework_application_events.yml
```diff
@@ -49,12 +49,22 @@
...
               ICON=:frame_with_picture:
               STAGE={{ environment }}
               STATUS=FAILED
               URL=<${BUILD_URL}consoleFull|#${BUILD_NUMBER}>
               CHANNEL=#dm-2ndline
+          - project: notify-slack
+            condition: SUCCESS
+            predefined-parameters: |
+              USERNAME=framework-application-event-email
+              JOB=Notify suppliers of new ${FRAMEWORK_SLUG} framework event - {{ environment }}
```
katstevens: It'd be great if we could include which event it was, e.g. `40_standstill_starts_today`

lfdebrux: How about adding it as a `PROJECT`?

katstevens: How does that appear in slack?

lfdebrux: The idea is that it would look something like this, but with the variable filled in

image

lfdebrux: It looks however like we're limited by whatever variables are defined by the job parameters :(

So either the job needs to be refactored or we have a super long text in the project section.

lfdebrux: Here's what it looks like with the super long email type

image

lfdebrux: image

https://gds.slack.com/archives/CAC985MCZ/p1584715643028300

lfdebrux:
These changes have been tested on Jenkins, you can see an example build here:

https://ci.marketplace.team/job/notify-suppliers-of-framework-application-event-preview/11/console

job_definitions/notify_suppliers_of_framework_application_events.yml
```diff
@@ -65,7 +80,8 @@
             FLAGS="$FLAGS --dry-run"
           fi
 
-          NOTIFY_TEMPLATE_ID="$(echo $EMAIL_TYPE | cut -d ':' -f 2 -)"
+          NOTIFY_TEMPLATE_MAP='{{ notify_template_map | to_json }}'
+          NOTIFY_TEMPLATE_ID="$(echo $NOTIFY_TEMPLATE_MAP | jq '.[env.EMAIL_TYPE]')"
```
katstevens: `jq` to the rescue! Good spot

PR #315, opened 2020-05-12
Allow JavaScript eval in HTML published by Jenkins

lfdebrux:
Fixes https://trello.com/c/NtZUnaSI

tl;dr:
Since updating BackstopJS the HTML reports have been doing something not
allowed by Jenkins' Content Secuirty Policy; this commit pokes a hole to
allow the bad behaviour.

### DIFF
--- a/playbooks/roles/jenkins/templates/jenkins_defaults.j2
+++ b/playbooks/roles/jenkins/templates/jenkins_defaults.j2
@@ -11,7 +11,7 @@ JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF8
 JAVA_ARGS="-Djava.awt.headless=true -Dfile.encoding=UTF8 -Xmx4096m -XX:MaxPermSize=1024m"  # Allow graphs etc. to work even when an X server is present
 
 # Allow JS and CSS on HTML Publisher pages
-JAVA_ARGS="${JAVA_ARGS} -Dhudson.model.DirectoryBrowserSupport.CSP=\"sandbox allow-scripts; default-src 'self' 'unsafe-inline' data:;\""
+JAVA_ARGS="${JAVA_ARGS} -Dhudson.model.DirectoryBrowserSupport.CSP=\"sandbox allow-scripts; default-src 'self' 'unsafe-inline' 'unsafe-eval' data:;\""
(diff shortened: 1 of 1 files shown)


playbooks/roles/jenkins/templates/jenkins_defaults.j2
```diff
@@ -11,7 +11,7 @@ JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF8
 JAVA_ARGS="-Djava.awt.headless=true -Dfile.encoding=UTF8 -Xmx4096m -XX:MaxPermSize=1024m"  # Allow graphs etc. to work even when an X server is present
 
 # Allow JS and CSS on HTML Publisher pages
-JAVA_ARGS="${JAVA_ARGS} -Dhudson.model.DirectoryBrowserSupport.CSP=\"sandbox allow-scripts; default-src 'self' 'unsafe-inline' data:;\""
+JAVA_ARGS="${JAVA_ARGS} -Dhudson.model.DirectoryBrowserSupport.CSP=\"sandbox allow-scripts; default-src 'self' 'unsafe-inline' 'unsafe-eval' data:;\""
```
katstevens, line 14: Does this mean we'll have a new warning in the Jenkins UI to dismiss? 😹

lfdebrux: I don't think so? Jenkins is currently running with this setting and I haven't seen a warning.

katstevens: That's because they only show up when we restart Jenkins (and we then dismiss the warnings...).

lfdebrux: Right, of course. I can apply this and see whether it will cause an error? I'm not sure what we could do if it did though. Alternatively we can rollback the changes to BackstopJS until they remove the `eval` from their code, but I don't think that's currently on their roadmap.

katstevens: I think it's fine for now. Long term we might be thinking about alternatives to BackstopJS anyway...

PR #319, opened 2020-05-28
Allow AWS CLI to switch infrastructure roles

katstevens:
https://trello.com/c/kCUjDnH4/48-3-3-jenkins-job-for-publish-services

The Jenkins user already has permission to assume the infrastructure roles, but the AWS CLI client needs to be configured on the Jenkins instance to actually do it.

The variables `cloudtrail_validate_logs_roles` and `aws_infrastructure_roles` are extracted from the credentials repo [when the Ansible playbook is initialised](https://github.com/alphagov/digitalmarketplace-jenkins/blob/master/Makefile#L50).  As such, the [PR to add the new var](https://github.com/alphagov/digitalmarketplace-credentials/pull/321) will need to be merged before this one.

### DIFF
--- a/playbooks/roles/jenkins/templates/aws_config.j2
+++ b/playbooks/roles/jenkins/templates/aws_config.j2
@@ -1,3 +1,5 @@
+[default]
+region=eu-west-1
 {% for role in assume_cloudtrail_roles %}
 
 [profile {{role.profile.name}}]
--- a/playbooks/roles/jenkins/templates/aws_config.j2
+++ b/playbooks/roles/jenkins/templates/aws_config.j2
@@ -5,3 +5,9 @@ credential_source=Ec2InstanceMetadata
 region={{role.profile.region}}
 role_arn={{role.profile.role_arn}}
 {% endfor %}
+{% for role in aws_infrastructure_roles %}
+
+[profile {{role.profile.name}}]
+credential_source=Ec2InstanceMetadata
+role_arn={{role.profile.role_arn}}
+{% endfor %}
--- a/playbooks/roles/jenkins/templates/aws_config.j2
+++ b/playbooks/roles/jenkins/templates/aws_config.j2
@@ -1,4 +1,4 @@
-{% for role in assume_roles %}
+{% for role in assume_cloudtrail_roles %}
 
 [profile {{role.profile.name}}]
 credential_source=Ec2InstanceMetadata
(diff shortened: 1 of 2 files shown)


playbooks/roles/jenkins/templates/aws_config.j2
```diff
@@ -1,7 +1,13 @@
...
 [profile {{role.profile.name}}]
 credential_source=Ec2InstanceMetadata
 region={{role.profile.region}}
 role_arn={{role.profile.role_arn}}
 {% endfor %}
+{% for role in aws_infrastructure_roles %}
+
+[profile {{role.profile.name}}]
+credential_source=Ec2InstanceMetadata
+role_arn={{role.profile.role_arn}}
```
lfdebrux, line 12: Why isn't the region required for these? 🤔

katstevens: I was hoping someone else could answer that 😹  I'm not 100% sure if it's needed for the Cloudtrail profiles!

katstevens: In practice though, the Jenkins user is only going to be using `aws s3` commands, so they should be region agnostic.

lfdebrux: If we can take the regions out for the cloudtrail profiles as well that would be great!

lfdebrux: Maybe what we can do is add

```
[default]
region=eu-west-1
```

to the top of the file, then we don't need regions for the individual profiles, and we can use one for loop for all of the profiles.

katstevens: Yeah, I think adding a default region is a good shout.

I think I'd rather keep the loops separate though as they're coming from different creds variables?

lfdebrux: But then we'd have to add new loops every time we want to add more creds variables? I guess that's not a big deal really

katstevens: Given the internal structure of the vars is not very visible, I think some verbosity here is fine.

katstevens: I've pushed a commit to add the default region. However as Cloudtrail is region-specific I'm going to keep the explicit setting for each profile there for now.

PR #321, opened 2020-06-01
Downgrade Ansible

katstevens:
Snyk overeagerly upgrade Ansible for us - [the version is no longer available to download](https://pypi.org/project/ansible/#history). Let's revert back to the previous version.

To prevent this happening again, I've added a Travis config file to try and install the requirements before merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: python
+python:
+  - "3.6"
+  - "3.7"
+  - "3.8"
+install:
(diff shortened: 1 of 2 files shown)


katstevens:
Now, how can we convince Snyk that the remediation is no longer available?

lfdebrux:
> Now, how can we convince Snyk that the remediation is no longer available?

I'm not sure...

PR #322, opened 2020-06-01
Jenkins job for publish-draft-services script

katstevens:
https://trello.com/c/kCUjDnH4/48-3-3-jenkins-job-for-publish-services

This job should be able to switch roles using the new AWS profile configuration.

I've specified the `docker run` [volume flag](https://docs.docker.com/engine/reference/run/#volume-shared-filesystems) `-v $HOME/.aws:/root/.aws` as per the [validate cloudtrail logs job](https://github.com/alphagov/digitalmarketplace-jenkins/blob/master/job_definitions/validate_cloudtrail_logs.yml#L30).

The `draft-ids` parameter has been left out of the script for now, as it refers to a file upload. That parameter is likely to be used on a failure re-run (with a small number of IDs) and should probably be handled manually rather than on Jenkins anyway.

### DIFF
--- a/job_definitions/publish_draft_services.yml
+++ b/job_definitions/publish_draft_services.yml
@@ -1,0 +1,74 @@
+{% for environment in ['preview', 'production'] %}
+- job:
+    name: "publish-draft-services-{{ environment }}"
+    display-name: "Publish draft services - {{ environment }}"
+    project-type: freestyle
+    disabled: true
+    description: |
+      This job runs the publish-draft-services script, for a 'pending' or 'standstill' framework.
+
+      For each draft service submitted by a successful supplier, a new 'live' service object is created via the API.
+      The draft service is updated with the new service ID. For G-Cloud only, any documents associated with the service
+      are then copied to a public S3 bucket.
+
+      Warning: this job takes approximately 10 hours to run for a G-Cloud framework with 36,000 draft services.
+
+    parameters:
+      - string:
+          name: FRAMEWORK_SLUG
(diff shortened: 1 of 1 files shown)


job_definitions/publish_draft_services.yml
```diff
@@ -0,0 +1,74 @@
...
+      - bool:
+          name: SKIP_DOCS_IF_PUBLISHED
+          default: false
+          description: "Don't copy documents if the draft service has already been published."
+      - bool:
+          name: DRY_RUN
+          default: false
+          description: "Dry run - do not set values in the database or copy documents"
+    publishers:
+      - trigger-parameterized-builds:
```
lfdebrux, line 33: Can we also have a Slack message when this job starts?

katstevens: I am a little wary of adding more noise to the channel, but if it will be useful for people then happy to add?

lfdebrux: My reasoning is that we're going to be running this job very rarely, so we're not going to be seeing the message very often, and it might be useful for people to get a Slack reminder that this is going on. Obviously we could just post a message ourselves though.

katstevens: Yeah, I think a human posting an `@here` is probably more visible than a slackbot. Unless I get really creative with the icon 😹

katstevens:
Going to merge and apply this for now - unfortunately it's annoying to test without actually being the `jenkins` user inside the box (to access the EC2 creds that will allow the jenkins user to switch roles). So I expect to have to tweak this in another PR.

PR #325, opened 2020-06-22
Enforce AWS_PROFILE choices on publish-draft-services

katstevens:
https://trello.com/c/ui2WdVFw/148-05-make-publish-draft-services-jenkins-job-dos-friendly

I did a quick local test for DOS, and while the AWS_PROFILE isn't needed for that framework, if a malformed value is provided (e.g. `bananas-infrastructure`) then the script will fail. So I've enforced the available values as choices on the job and clarified that either value can be used for DOS.

### DIFF
--- a/job_definitions/publish_draft_services.yml
+++ b/job_definitions/publish_draft_services.yml
@@ -21,10 +21,17 @@
       - string:
           name: FRAMEWORK_SLUG
           description: "The framework slug to publish services for, e.g. 'g-cloud-11'."
-      - string:
+      - choice:
           name: AWS_PROFILE
-          default: "{% if environment == 'production' %}production{% else %}development{% endif %}"
-          description: "AWS profile to copy documents as. Only used for G-Cloud frameworks."
+          choices:
+{% if environment == 'production' %}
+            - production
+            - development
+{% else %}
+            - development
+            - production
+{% endif %}
+          description: "AWS profile to copy documents as. Only used for G-Cloud frameworks - either value can be entered for DOS frameworks."
       - bool:
           name: SKIP_DOCS_IF_PUBLISHED
           default: false


lfdebrux:
Does an empty value cause the script to crash? Then we could have the empty string as the first choice?

katstevens:
> Does an empty value cause the script to crash for DOS frameworks? Then we could have the empty string as the first choice?

We have to put something, otherwise we get: `The config profile () could not be found`

job_definitions/publish_draft_services.yml
```diff
@@ -23,8 +23,11 @@
           description: "The framework slug to publish services for, e.g. 'g-cloud-11'."
       - string:
           name: AWS_PROFILE
+          choices:
+            - production
+            - development
           default: "{% if environment == 'production' %}production{% else %}development{% endif %}"
```
lfdebrux, line 29: I think choice parameters use a different syntax in Jenkins Job Builder ([docs](https://docs.openstack.org/infra/jenkins-job-builder/parameters.html#parameters.choice)).

The key needs to be `choice`, and I don't know if the value of the `default` key is respected, the docs just say that the first in the list is the default 🤔 .

```suggestion
      - choice:
          name: AWS_PROFILE
          choices:
            - production
            - development
          default: "{% if environment == 'production' %}production{% else %}development{% endif %}"
```

katstevens: Ooo good spot. If there's no default wonder if I can do

```
- choice:
  name: AWS_PROFILE
  choices:
{% if environment == 'production' %}
     - production
     - development
{% else %}
     - development
     - production
{% endif %}
```

lfdebrux: That should work!

PR #329, opened 2020-07-07
Remove jobs notify-suppliers-of-framework-clarification-questions-answers

lfdebrux:
We have two different Jenkins jobs that can send email 21 "New framework clarification questions and answers have been uploaded". This is because we intended to replace the job notify-suppliers-of-framework-clarification -questions-answers with the more generic and capable Jenkins job notify -suppliers-of-framework-application-event, but we never got around to it (see alphagov/digitalmarketplace-scripts#475 for the history of this).

However, having the two different jobs has caused some confusion, and we've been happily using notify-suppliers-of-framework-application-event for some time, so we can get rid of the older job.


katstevens:
Can we [remove the job from the tabs](https://github.com/alphagov/digitalmarketplace-jenkins/blob/master/playbooks/roles/jenkins/defaults/main.yml#L209) as well please?

lfdebrux:
> Can we [remove the job from the tabs](https://github.com/alphagov/digitalmarketplace-jenkins/blob/master/playbooks/roles/jenkins/defaults/main.yml#L209) as well please?

Yep, of course, I always forget to do that, thanks.

PR #337, opened 2020-08-18
Update Content Security Policy to allow-same-origin

katstevens:
https://trello.com/c/sXG02trb/1839-visual-regression-test-report-403

We can't view HTML reports (which use iframes) for our Visual Regression tests on Jenkins, due to changes in how Chrome handles behaviour for cookies without a `SameSite` attribute.

This is a quick fix while we investigate longer term options.


lfdebrux:
We can test this fairly straightforwardly by changing the CSP on Jenkins manually: https://wiki.jenkins.io/display/JENKINS/Configuring+Content+Security+Policy#ConfiguringContentSecurityPolicy-Implementation

katstevens:
It works! Thanks @lfdebrux

PR #340, opened 2020-09-22
Update mark-definite-framework-results to log Jenkins user in audit logs

lfdebrux:
We want the Jenkins user who starts the mark-definite-framework-results to be recorded in our audit log, so we can see later who made a change.

By default the mark-definite-framework-results script uses the unix username, which on Jenkins is not very useful. Instead we use the build tag and build user envvars from the [build user vars plugin](https://plugins.jenkins.io/build-user-vars-plugin/) to make a more useful message.


lfdebrux:
I've tested this on Jenkins

bjgill (approved):
Looks sensible - and much more readable on multiple lines.

PR #343, opened 2020-09-23
Add system dependencies for new counterpart signature PDF script

lfdebrux:
https://trello.com/c/4Su1kLUN/397-update-apt-get-dependencies-for-e-signature-counterpart-agreement-script

We need the official build of wkhtmltopdf with patched QT to be able to get page numbers working in the latest agreement PDF format. Also, to make sure the fonts are consistent across the document we need Microsoft core fonts installed.

### DIFF
--- a/playbooks/roles/jenkins/defaults/main.yml
+++ b/playbooks/roles/jenkins/defaults/main.yml
@@ -35,6 +35,7 @@ dist_tools:
   - python3-dev
   - python3-pip
   - python3-venv
+  - ttf-mscorefonts-installer
   - unzip
   - xvfb
   - zip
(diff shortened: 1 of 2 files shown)


gidsg:
I think we may need to do some extra steps to accept the Microsoft EULA i.e. https://askubuntu.com/a/25614/234184 ?

lfdebrux:
> I think we may need to do some extra steps to accept the Microsoft EULA i.e. https://askubuntu.com/a/25614/234184 ?

Ah good point.

There's an example of how to do it in ansible here https://github.com/pludoni/ansible-wkhtmltox/blob/master/tasks/font.yml

lfdebrux:
> I think we may need to do some extra steps to accept the Microsoft EULA i.e. https://askubuntu.com/a/25614/234184 ?

Added.

gidsg (approved):
🔡 👍

PR #344, opened 2020-09-24
Fix typo in publish draft services script

lfdebrux:
Missing a dollar sign that meant --skip-docs-if-published flag was never added to $FLAGS, meaning that the build would take way longer than needed if restarting a build failed halfway-through.


domoscargin (approved):
LGTM

PR #355, opened 2020-10-19
Move config for Jenkins views to Jenkins Job Builder

lfdebrux:
https://trello.com/c/GXlRVRtN/980-1dmanage-our-jenkins-list-views-using-jenkins-job-builder

Use Jenkins Job Builder to manage Jenkins views, instead of generating the config with Jinja.

This has the advantage that running `make jenkins TAGS=jobs` will now update lists as well as jobs.

Jenkins Job Builder currently does not support the Build Monitor view plugin, so for now we leave that as is.


domoscargin (approved):
LGTM!

PR #361, opened 2020-11-13
Add notes on rotating secrets

lfdebrux:
I found the process of doing this a little surpising; I've added some notes I would have found useful to the job description for next time.

### DIFF
--- a/job_definitions/rotate_ft_account_passwords.yml
+++ b/job_definitions/rotate_ft_account_passwords.yml
@@ -29,8 +29,8 @@
           milestone()
           waitUntil {
             try {
-              input(message: "1/2: You should now temporarily disable the functional tests, smoke tests and smoulder tests for this stage")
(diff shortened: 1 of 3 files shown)


job_definitions/rotate_ft_account_passwords.yml
```diff
@@ -81,18 +81,4 @@
               scripts/rotate-api-tokens.sh sync-ft-account-passwords $STAGE
           ''')
         }
-        stage("Re-enable functional/smoke/smoulder tests") {
```
bjgill, line 84: Is `make jenkins TAGS=jobs JOBS=smoke_smoulder_tests` the thing that makes this redundant?

bjgill: Never mind - just seen the linked ticket in the commit description. Makes sense now.

bjgill (approved):
:book: :tada:

PR #363, opened 2020-11-19
Change update index job to use parallel indexing by default

lfdebrux:
This should speed up indexing considerably.


alex9smith (approved):
🕙

PR #364, opened 2020-11-20
Disable regular database backups

bjgill:
Our database has now grown too big for this job: `OverflowError: string longer than 2147483647 bytes`

So we're going to disable it until we know what we want to do. We'll rely on the PaaS backups in the mean time.


gidsg (approved):
This is OK for now, although we do need a proper solution as I believe these dumps are also used for the cleaned snapshots on other environments.

PR #365, opened 2020-12-03
Update playbook and add instructions for creating a new Jenkins instance

alex9smith:
While setting up a [new Jenkins instance for the ITHC](https://trello.com/c/nCAuU2nf/558-3-give-the-pentester-a-way-to-test-our-jenkins-instance) I had to update a couple of things in the playbook to get it to work:

1. The official Jenkins apt repository has moved
2. We now need to add `matrix-auth` to the list of plugins.

I've also added instructions to the README covering how to set up a new instance, as there are some things which need to happen outside this repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,10 @@ the `dm-ssp-jenkins` user. The Client ID and Client Secret must be stored in the
 *jenkins-vars/jenkins.yaml*, and must be stored as a nested dict under `jenkins_github_auth_by_hostname`. This allows
 us to maintain multiple Jenkins instances (if required). These credentials are deployed via the `config` task tag.
 
+## Creating a new Jenkins instance
+
+The process for creating a new instance is documented in [the team manual](https://alphagov.github.io/digitalmarketplace-manual/2nd-line-runbook/rebuilding-jenkins.html).
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -122,6 +122,17 @@ the `dm-ssp-jenkins` user. The Client ID and Client Secret must be stored in the
 *jenkins-vars/jenkins.yaml*, and must be stored as a nested dict under `jenkins_github_auth_by_hostname`. This allows
 us to maintain multiple Jenkins instances (if required). These credentials are deployed via the `config` task tag.
 
+## Creating a new Jenkins instance
```
lfdebrux, line 125: Is this all the info that was needed to create a Jenkins instance? You didn't need to follow any of the stuff in the manual?

alex9smith: Oh of course... I was going to link to that manual page here and then completely forgot

alex9smith: Actually, I'm just going to replace what I wrote here with a link to the manual. It's got far more detail and there's no point duplicating the documentation

bjgill (approved):
Laurence is off today, but this looks good to me.

PR #375, opened 2021-01-15
Add --rm to all Docker run commands

lfdebrux:
We have a lot of dangling containers on our Jenkins instance, we should make sure to clean them up when they are not used. Adding `--rm` to everywhere we use `docker run` is a quick fix.


bjgill (approved):
:fire_extinguisher:

PR #386, opened 2021-02-16
Add main to branches to backup for code backups

lfdebrux:
Ticket: https://trello.com/c/DrVWMkcZ/2169-rename-default-branches-on-our-repos-from-master-to-main

We are renaming our default branches in our git repos to `main`, so we want to back those up as well as/instead of branches named `master`.

While we are in the process of migrating (which could take a while), we just cover all our bases.

### DIFF
--- a/job_definitions/code_backups.yml
+++ b/job_definitions/code_backups.yml
@@ -44,6 +44,7 @@
           credentials-id: github_com_and_enterprise
           branches:
             - master
+            - main
           wipe-workspace: false
     properties:
       - build-discarder:


job_definitions/code_backups.yml
```diff
@@ -44,6 +44,7 @@
           credentials-id: github_com_and_enterprise
           branches:
             - master
+            - main
```
bjgill, line 47: What happens for those repos where the `main` branch has not yet been created?

lfdebrux: Good question... I'm going to apply the change and find out!

lfdebrux: I've applied the changes to Jenkins with `make jenkins TAGS=jobs`, and repos which don't have a `main` branch yet work fine still: here's the [git polling log for digitalmarketplace-utils](https://ci.marketplace.team/view/Code%20backups/job/backup-digitalmarketplace-utils/scmPollLog/) as an example.

However, repos which have had their default branch renamed are not so happy :( the [most recent backup of digitalmarketplace-test-utils](https://ci.marketplace.team/view/Code%20backups/job/backup-digitalmarketplace-test-utils/503/console) failed with the error `remote: error: refusing to delete the current default branch 'refs/heads/master'.` 😢

bjgill: Hmm. Do we just need to SSH into Jenkins and clear the cached workspace for this job? I think that might solve the error...

lfdebrux: I will do some experiments...

lfdebrux: I think this might be more to do with what's happening on AWS CodeCommit, based on this stack exchange thread

https://superuser.com/questions/748058/how-do-i-delete-origin-master-in-git

lfdebrux:
Putting this down while I try to see how we can change the checked out branch in AWS CodeCommit, also I need to work on other things

gidsg:
@lfdebrux I think this should work in conjunction with updating the CodeCommit default branch as described in https://docs.aws.amazon.com/codecommit/latest/userguide/how-to-change-branch.html as such I think it should be ready to merge.

gidsg (approved):
👍 🍴

PR #387, opened 2021-02-18
delete code before backup

gidsg:
https://trello.com/c/DrVWMkcZ/2169-rename-default-branches-on-our-repos-from-master-to-main

If a previous job fails it can leave a non-empty folder which causes subsequent jobs to fail:
e.g.
```
git clone https://4c3123d80a3e5000b9ed2ae0556f2dfb026d17b4@github.com/alphagov/digitalmarketplace-runner digitalmarketplace-runner --mirror
fatal: destination path 'digitalmarketplace-runner' already exists and is not an empty directory.
```
To fix this we delete the folder in case it exists when the job starts.

### DIFF
--- a/job_definitions/code_backups.yml
+++ b/job_definitions/code_backups.yml
@@ -57,6 +57,7 @@
       - ansicolor
     builders:
       - shell: |
+          rm -rf {{ package }}
           git clone https://$GITHUB_ACCESS_TOKEN@github.com/alphagov/{{ package }} {{ package }} --mirror
           cd {{ package }}
           git remote add {{ package }}-aws-backup-remote https://git-codecommit.eu-west-1.amazonaws.com/v1/repos/{{ package }}


job_definitions/code_backups.yml
```diff
@@ -57,6 +57,7 @@
       - ansicolor
     builders:
       - shell: |
+          rm -rf {{ package }}
```
bjgill, line 60: From line 47 above:

```
          wipe-workspace: false
```

Would setting this to `true` have the same effect as the `rm -rf`?

bjgill (approved):
I assume this was the thing you were talking about that can be removed once we've migrated all our repos to `main`?

job_definitions/code_backups.yml
gidsg: Ah, good point, I missed that flag. I can try and set it and see how it behaves on Jenkins

gidsg:
@bjgill, no the temporary thing is running the job and then manually changing the default branch in CodeCommit. This should remain as it helps to have a clean state if it fails part-way for whatever reason.

job_definitions/code_backups.yml
```diff
@@ -57,6 +58,7 @@
       - ansicolor
     builders:
       - shell: |
+          cd ${WORKSPACE}
```
bjgill, line 61: Why is this necessary?

gidsg: Sorry, I pushed by accident whilst I was testing different approaches. I will force push once I have a better solution.

bjgill: OK - let me know when you want me to take a proper look.

gidsg:
@bjgill I tried `wipe-workspace: true` however, it did not appear to have the intended effect, i.e. the directory still exists e.g. in https://ci.marketplace.team/view/Code%20backups/job/backup-digitalmarketplace-runner/562/console so I have reverted to the previous approach.

gidsg:
@bjgill as discussed, I have force-pushed to provide more details in the commit extended description.

bjgill (approved):
:shipit:

PR #398, opened 2021-03-15
Update digitalmarketplace-scripts jobs to use invoke instead of make

lfdebrux:
alphagov/digitalmarketplace-scripts#649 replaced make with invoke in the scripts repo, we need to update our job definitions to reflect this.

### DIFF
--- a/job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
+++ b/job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
@@ -64,10 +64,9 @@
                   git clean -fdx
                   git reset --hard origin/master
 
-                  make virtualenv
-                  ./venv/bin/pip install --upgrade pip
-                  ./venv/bin/pip install wheel  # this allows pip to cache compiled packages instead of rebuilding them all on every invocation
-                  make requirements
+                  pip install -U digitalmarketplace-developer-tools
+                  invoke virtualenv
+                  invoke requirements
                 ''')
               }
             }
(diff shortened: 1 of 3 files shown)


job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
```diff
@@ -64,18 +64,17 @@
...
+                  invoke requirements
                 ''')
               }
             }
             stage('Generate') {
               dir('digitalmarketplace-scripts') {
                 sh('''
                   [ "$DRY_RUN" = "true" ] && FLAGS="$FLAGS --dry-run"
-                  
+
```
gidsg, line 77: very tiny but is this whitespace change intended?

gidsg (approved):
Looks fine to me, have you tested on Jenkins?

job_definitions/generate_upload_and_notify_counterpart_signature_pages.yml
lfdebrux: Yes, my editor was complaining about it.

lfdebrux:
@gidsg tested it just now, it turns out the pip bin directory isn't in the Jenkins user's path, so I need to make a few more changes.

lfdebrux:
I've pushed some changes to make ~jenkins/.local/bin available in PATH, so this works now:

https://ci.marketplace.team/job/generate-upload-and-notify-counterpart-signature-pages/1461/console

PR #404, opened 2021-03-22
Change build-image parameter REBUILD to default true

lfdebrux:
We want to be able to rebuild Docker images for apps after base images have been rebuilt, so we can get the latest version of security patches.  This commit changes the default of the build-image job so that it will not skip the Docker build if the release tag for the app image already exists on Docker Hub. Because the Docker client is already pretty good at caching build products this shouldn't result in much wasted work.


gidsg (approved):
👍

PR #405, opened 2021-03-23
Make common tags into their own targets

bjgill:
https://trello.com/c/03kE0zPn/858-05-create-more-explicit-make-targets-in-digitalmarketplace-jenkins

We recently had [an incident](https://docs.google.com/document/d/15rVb16Nex37ZF8DJsLBSUcAeYxNPlWabxr4c6feRwO0) during which a developer accidentally ran `make jenkins` using the wrong tag and caused Jenkins to restart at an inopportune moment.

Add targets for the most common tags. This should make it a bit more difficult to do the wrong thing in future.

Leave the existing target so that people can still choose to run uncommon tags. This avoids us needing to maintain a vast list of rarely-used targets.

We'll need to update the team manual after this gets merged.

Also, update `jobs` when we update `config` to fix https://trello.com/c/PWoPSBNv

### DIFF
--- a/deploy-jenkins.sh
+++ b/deploy-jenkins.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/bash
+set -eou pipefail
+# Script to run ansible tasks. Use the Makefile instead of running this directly.
+
+${DM_CREDENTIALS_REPO}/sops-wrapper -v > /dev/null
+
+JENKINS_VARS_FILE=$(mktemp)
+PRIVATE_KEY_FILE=$(mktemp)
+trap 'rm $JENKINS_VARS_FILE $PRIVATE_KEY_FILE' EXIT
+
+for varfile in ${DM_CREDENTIALS_REPO}/jenkins-vars/*.yaml
+do
+  ${DM_CREDENTIALS_REPO}/sops-wrapper -d $varfile >> $JENKINS_VARS_FILE
+done
+${DM_CREDENTIALS_REPO}/sops-wrapper -d ${DM_CREDENTIALS_REPO}/aws-keys/ci.pem.enc > $PRIVATE_KEY_FILE
+
+echo ${VIRTUALENV_ROOT}/bin/ansible-playbook \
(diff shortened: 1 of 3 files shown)


deploy-jenkins.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/bash
```
lfdebrux, line 1: 🙌 great idea

deploy-jenkins.sh
```diff
@@ -0,0 +1,23 @@
...
+done
+${DM_CREDENTIALS_REPO}/sops-wrapper -d ${DM_CREDENTIALS_REPO}/aws-keys/ci.pem.enc > $PRIVATE_KEY_FILE
+
+echo ${VIRTUALENV_ROOT}/bin/ansible-playbook \
+  -i playbooks/hosts playbooks/jenkins_playbook.yml \
+  -e @$JENKINS_VARS_FILE \
+  -e "jenkins_public_key='$(ssh-keygen -y -f $PRIVATE_KEY_FILE)'" \
+  --key-file=$PRIVATE_KEY_FILE \
+  -e "dm_credentials_repo=${DM_CREDENTIALS_REPO}" \
+  --tags "${TAGS}" ${EXTRA_VARS:-}
```
lfdebrux, line 23: Why is the `${parameter:-word}` expansion required here?

lfdebrux (approved):

deploy-jenkins.sh
bjgill: I've used the [bash 'strict mode'](http://redsymbol.net/articles/unofficial-bash-strict-mode/).

`set -u` means that undefined variables become errors. `EXTRA_VARS` is the one variable where it's OK for it to be undefined. So I added that expansion so that it wouldn't get picked up by `set -u`

lfdebrux: Ooh I didn't know about `-u`, thanks!

PR #409, opened 2021-03-25
Move extra variable calculation into bash

bjgill:
https://trello.com/c/03kE0zPn/858-05-create-more-explicit-make-targets-in-digitalmarketplace-jenkins

Previously, we constructed `EXTRA_VARS` in make, which bash interpreted (and passed to ansible) as a single string. Ansible did not recognise this as a valid extra var.

Now, we construct `EXTRA_VARS` in bash, where we can use a bash array to ensure that it all works correctly.

Also replace short options with long options for legibility.

Fixes the final bug I introduced in https://github.com/alphagov/digitalmarketplace-jenkins/pull/405. We've now tested that this works by running it against the real Jenkins (from both MacOS and Ubuntu).


gidsg (approved):
I have tested this locally and it works as expected 👍

PR #410, opened 2021-04-01
Remove job to send stats to performance platform

bjgill:
The Performance Platform has passed on. It is no more. It has ceased to be.


lfdebrux (approved):
It's not dead it's resting...

PR #426, opened 2021-05-24
Notify Slack if the code backup job fails

alex9smith:
Previously we didn't get a notification in Slack if the code backup jobs failed. This could cause problems if we needed to use them and found out at that point that they weren't working.

Uses a [publisher](https://docs.openstack.org/infra/jenkins-job-builder/publishers.html) to trigger an action after the build completes. See the [`export_supplier_data_to_s3`](https://github.com/alphagov/digitalmarketplace-jenkins/blob/main/job_definitions/export_supplier_data_to_s3.yml) job for an example of how we already use this syntax.

https://trello.com/c/07xZl8W7/973-1-we-dont-get-an-alert-when-our-code-backup-jobs-fail

### DIFF
--- a/job_definitions/code_backups.yml
+++ b/job_definitions/code_backups.yml
@@ -57,6 +57,18 @@
           ignore-post-commit-hooks: True
     wrappers:
       - ansicolor
+    publishers:
+      - trigger-parameterized-builds:
+          - project: notify-slack
+            condition: UNSTABLE_OR_WORSE
+            predefined-parameters: |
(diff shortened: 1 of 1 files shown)


gidsg (approved):
👍  LGTM, worth testing after this deployed it is working as expected (if you are able to induce a failure?)

alex9smith:
I think I can deploy it with an extra non-existent package name to check the message gets sent, then revert back once we know it's working

PR #429, opened 2021-06-03
Throttle app pipelines to 4 concurrent jobs

bjgill:
https://trello.com/c/FA715OCK/1566-prevent-jenkins-deadlocks

We just had Jenkins deadlock. This is because we had a large number of app pipeline jobs all in progress. However, they were all blocked on a deploy job, so could not make progress. However, all the deploy jobs were blocked on other jobs, mainly notifying Slack. However, the jobs in this third tier couldn't be scheduled because all the executors were full of blocked jobs in the first and second tier. We had to cancel a bunch of jobs to clear the deadlock.

Introduce [throttling](https://plugins.jenkins.io/throttle-concurrents/#throttling-of-pipeline-jobs) on the app pipelines. Now, no more than 4 can run at once. This means that in the worst case, only 12 executor slots will be occupied by jobs related to app pipelines. So other jobs will still be able to run. I think this should make it much harder to accidentally deadlock Jenkins, and allow us to merge with impunity.

We previously tried doing this in https://github.com/alphagov/digitalmarketplace-jenkins/pull/400. That didn't work because the throttling was done on too wide a scope, so picked up pipelines that weren't even trying to make progress. I think this approach should work, by only throttling just before we want to allocate an executor node.

Let's see if this works, and quickly revert if not.

### DIFF
--- a/job_definitions/release_pipeline.yml
+++ b/job_definitions/release_pipeline.yml
@@ -27,9 +27,11 @@
       def run_functional_tests(stage) {
         waitUntil {
           try {
-            node {
-              build job: "functional-tests-${stage}"
-              return true
+            throttle(['AppPipelines']) {
+              node {
+                build job: "functional-tests-${stage}"
+                return true
+              }
             }
           } catch(error) {
             echo "Functional tests failed: ${error}"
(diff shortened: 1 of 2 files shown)


gidsg (approved):
Let's give this a go! 👍

PR #432, opened 2021-06-11
Update to Ansible 4

bjgill:
We can't simply upgrade from ansible 3 to 4 with pip - it doesn't work.

So detect whether the venv has got ansible 4 and uninstall ansible if not. This will allow the upgrade to ansible 4 to happen automatically without errors.


gidsg (approved):
👍

PR #433, opened 2021-07-01
Run functional tests using headless Firefox

bjgill:
Our tests are currently rather unreliable. I have a hypothesis that this might be PhantomJS-related and that switching to Selenium might improve things.

Do an experiment to find out whether this is the case. Build on https://github.com/alphagov/digitalmarketplace-functional-tests/pull/901 to switch the functional test jobs over to use Selenium with headless firefox. Install firefox so that this works.

My plan is to deploy it and see what happens. Should be quick and easy to revert if it breaks things.

Also, detect-secrets picked up some old unrelated changes.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "aws-auth.nix|tests/jenkins-vars.yml",
     "lines": null
   },
-  "generated_at": "2020-10-08T12:52:30Z",
+  "generated_at": "2021-07-01T14:18:43Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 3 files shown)


gidsg:
@bjgill I think the dependency is `geckodriver` as well as `firefox`? https://github.com/alphagov/digitalmarketplace-functional-tests/blob/b8fa5302f8ac8de6c0b53e76965c8626b085989b/README.md#L74

bjgill:
Yes - discovered that the hard way whilst experimenting yesterday. What I did was manually apply the changes to Jenkins, run the tests, and then undo the changes.

Lots of test failures, but they looked like incompatibilities between Selenium and PhantomJS rather than anything more serious. No obvious flakiness, so I think this might be a useful approach.

Closing this because I think we want to do it in another way (from my local testing, Chrome is faster than Firefox in running the tests).

PR #435, opened 2021-07-08
Stop changing the jenkins user's primary group

bjgill:
We create the user in bootstrap, where we (correctly) set jenkins as the primary group. However, it looks as if we were incorrectly then changing the primary group to docker by using 'group' rather than 'groups' - see https://docs.ansible.com/ansible/latest/collections/ansible/builtin/user_module.html#parameter-group

Change to use the correct parameter name, and also switch to the new way of providing parameters.


gidsg (approved):
👍 👤

PR #437, opened 2021-07-23
Install chromedriver

bjgill:
https://trello.com/c/b6owLEB0/75-3-get-all-functional-tests-working-in-chrome

Our tests are currently rather unreliable. We think that Selenium/Chromium might be more reliable than Poltergeist/PhantomJS. At the very least, they're not obsolete, so should be easier to debug than our current setup.

Note that we're using chromium rather than chrome. I think this is safe - they're effectively the same thing. Chromium is just much easier to deploy.

Follows on from https://github.com/alphagov/digitalmarketplace-jenkins/pull/433 - Chromium doesn't seem to have the same performance problems we had with Firefox.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "aws-auth.nix|tests/jenkins-vars.yml",
     "lines": null
   },
-  "generated_at": "2020-10-08T12:52:30Z",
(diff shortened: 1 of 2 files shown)


gidsg (approved):
📦 👍

PR #440, opened 2021-07-29
remove phantomjs

gidsg:
since https://github.com/alphagov/digitalmarketplace-functional-tests/pull/912 we no longer depend on phantomjs for functional tests.


bjgill (approved):
I assume the overnight testing was successful and we're not at risk of needing to revert? If so, looks good

PR #443, opened 2021-08-02
Changes to support Ubuntu 20 LTS

gidsg:
https://trello.com/c/cacoWXHC/63-5-upgrade-the-version-of-python-on-jenkins-to-at-least-38

Restores changes from #439 plus additional fix to support `chromium-chromedriver`.


alex9smith (approved):
👍

PR #445, opened 2021-08-06
Change default scripts branch to main in signature job

alex9smith:
This must have been missed in #415. The `master` branch no longer exists so this job is now failing.


daniele-occhipinti (approved):
Good spot.

PR #448, opened 2021-08-19
Send all Slack notifications to #dm-2ndline

bjgill:
This is a temporary measure. Our CCS techops friends only have access to #dm-2ndline. So send all Slack notifications there so that they can see them. This will make that channel a lot more spammy, but I think that's an acceptable price to pay.

We should be able to revert this change once we've switched over to a CCS Slack (or equivalent) where we can use more than one channel.

### DIFF
--- a/job_definitions/build_image.yml
+++ b/job_definitions/build_image.yml
@@ -77,7 +77,7 @@
                       string(name: 'USERNAME', value: 'build-image'),
                       string(name: 'ICON', value: ':sad-docker:'),
                       string(name: 'JOB', value: "Build Docker image: ${REPOSITORY} - ${RELEASE_NAME}"),
-                      string(name: 'CHANNEL', value: "#dm-release"),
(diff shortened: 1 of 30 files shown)


daniele-occhipinti:
This is a good idea.
Would this be a good time to try to rename the #dm-2ndline channel to something that may be a better reflection of its use?

phantas (changes requested):
Missing jobs:

- [x] job_definitions/create_index.yml
- [ ] job_definitions/maintenance_mode.yml

bjgill:
> This is a good idea.
> Would this be a good time to try to rename the #dm-2ndline channel to something that may be a better reflection of its use? And the answer may be that it is not.

If we carry on for a long time with just the one channel, I agree that it would be worth renaming it. However, I'm hopeful that this will be a short-lived state of affairs, so it's probably not worth renaming.

bjgill:
Good catch. Have now fixed

phantas (approved):

